PR #450, opened 2011-11-09
Allow logging to a file

sj26:
Resque takes too long to boot within my app before writing the pidfile so process monitoring tools get confused if I simply background the process. Using `BACKGROUND` alleviates this but I lose logging. This patch adds `LOGFILE` for specifying a... well... log file.

This pull request is a stop-gap for us, but maybe should prompt some nicer logging in general?

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -251,12 +251,13 @@ but in the Resque workers it's fine.
 
 ### Logging
 
-Workers support basic logging to STDOUT. If you start them with the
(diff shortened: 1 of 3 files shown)


IevgenRagulin:
Is this issue closed because the issue has been fixed? I'm still facing this issue with the latest version of resque

IevgenRagulin:
I've just realized that I'm facing this issue with resque-scheduler, not with resque https://github.com/resque/resque-scheduler/issues/565

PR #1053, opened 2013-06-25
Ensure run_at is a string when sorting by it.

jonhyman:
When using Resque and Sidekiq at the same time, resque-web errors out if there are any Sidekiq workers processing jobs. It seems it is because "run_at" in Resque is an iso8601 string, whereas in Sidekiq it's a number. As such, #to_s the run_at to ensure that we're dealing with a string (especially since the nil case compares against empty string).

### DIFF
--- a/lib/resque/server/views/working.erb
+++ b/lib/resque/server/views/working.erb
@@ -49,7 +49,7 @@
     </tr>
     <% end %>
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
After far too long, I'm going to merge this change, because it's very minor. We may or may not change this behavior in the future, I don't know what @mperham does with Sidekiq these days.

PR #1144, opened 2013-10-29
Prevent a Resque::Failure from being created when re-queuing jobs

dwilkie:
I'm currently using the `on_failure_retry` hook to re-enqueue jobs like this:

```ruby
def on_failure_retry(exception, *args)
  if exception.is_a?(Resque::TermException)
    Resque.enqueue(self, *args)
  end
end
```
In these cases I don't want a new Resque::Failure to be created and shown in ResqueWeb

Borrowing from the idea for the before_hooks where you can raise a `Resque::DontPerform` exception to prevent a job from being run, it would would be nice if we could do something like this:

```ruby
def on_failure_retry(exception, *args)
  if exception.is_a?(Resque::TermException)
    Resque.enqueue(self, *args)
    raise Resque::DontFail
  end
end
```
This would prevent the code `Resque::Failure.create` from being called.

I think this approach is more flexible than automatically trying to rescue from `Resque::TermException` because the approach can be used for any job that has been requeued.

If people are happy with this solution I'll go ahead and attach a pull-request.

### DIFF
--- a/docs/HOOKS.md
+++ b/docs/HOOKS.md
@@ -141,6 +141,7 @@ Modules are even better because jobs can use many of them.
       def on_failure_retry(e, *args)
         Logger.info "Performing #{self} caused an exception (#{e}). Retrying..."
         Resque.enqueue self, *args
+        raise Resque::DontFail.new # Don't add the failed job to the Resque::Failure backend
       end
     end
 
--- a/docs/HOOKS.md
+++ b/docs/HOOKS.md
@@ -104,7 +104,8 @@ The available hooks are:
   `Resque::Failure` backend.
 
 * `on_failure`: Called with the exception and job args if any exception occurs
-  while performing the job (or hooks), this includes Resque::DirtyExit.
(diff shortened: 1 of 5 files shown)


dyatlov:
@dwilkie any news on that approach?
I'm interested in this as im using requeues and in case of temporary error on a service side my redis db gets quickly increased to several tens of megabytes full with failure logs.

steveklabnik:
Hey there!

Sorry that your PR has (probably) languished. Resque development has been on hold for a long time. That said, @hoffmanc and I (as well as others, hopefully!) are going to be getting back to it. Which brings me to this PR.

Resque 2.0 is basically on hold, forever. We'll be swapping the `1-x-stable` and `master` branches sometime next week. For more details, see here: http://words.steveklabnik.com/rescuing-resque-again

As such, `master` won't be taking any improvements at this time, as this redux is basically dying off. So, thank you for the pull request, but I'm going to give it a close. Sorry again about the wait. If you'd like to pitch in on the new stuff, we'd love to have you! Just let us know.

PR #1155, opened 2013-12-12
Avoid masking exceptions that cause unregistering

davetron5000:
When unregistering a worker, there are various calls to Redis to try to fail the job that's in progress, however if said job is failing due to an unhandled exception but we then get an exception accessing Redis itself, the original exception is raised, making it impossible to figure out what else might've gone wrong and how to fix it.

FWIW, I see this very frequently and the log message "Failed to start worker {ex.message}" is never rendered.

To show how masking the real exception is a problem, consider the following situation:

1. Queue worker to charge someone $50
2. Will contacting the gateway, we loose the network
3. Job fails with "Connection Timed Out"
4. Exception bubbles up and Resque unregisters the worker, with the timeout exception as the value `exception` passed to `unregister_worker`.
5. When checking to see if still processing a job, timeout to Redis (which you could imagine might happen if Redis is cloud-hosted and the network issue were on the network where workers were running)

In the current implementation, we don't see the timeout to our credit card processor, and we don't see what job was even running, so we can't event see that there was a problem charging someone $50

The change below would favor the original exception to be raised, instead of the Redis Timeout, in this case.

Let me know if there is a better way to do this.

### DIFF
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -135,7 +135,7 @@ describe "Resque::Worker" do
     assert_equal('StandardError', Resque::Failure.all['exception'])
   end
 
-  test "does not mask exception when timeout getting job metadata" do
+  it "does not mask exception when timeout getting job metadata" do
     job = Resque::Job.new(:jobs, {'class' => 'GoodJob', 'args' => "blah"})
     @worker.working_on(job)
     Resque.redis.stubs(:get).raises(Redis::CannotConnectError)
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -135,6 +135,26 @@ describe "Resque::Worker" do
     assert_equal('StandardError', Resque::Failure.all['exception'])
   end
 
+  test "does not mask exception when timeout getting job metadata" do
+    job = Resque::Job.new(:jobs, {'class' => 'GoodJob', 'args' => "blah"})
+    @worker.working_on(job)
+    Resque.redis.stubs(:get).raises(Redis::CannotConnectError)
+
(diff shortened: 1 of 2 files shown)


steveklabnik:
I am restarting the build to investigate the failures.

steveklabnik:
Failures are still happening.

@davetron5000 any interest in keeping this PR alive?

davetron5000:
Force-pushed after a rebase.  Let's see what happens

test/worker_test.rb
```diff
@@ -135,6 +135,26 @@ def self.perform
     assert_equal('StandardError', Resque::Failure.all['exception'])
   end
 
+  test "does not mask exception when timeout getting job metadata" do
```
steveklabnik: i think `test` needs to be `it` here, that's causing the travis failure

steveklabnik:
Thanks for keeping up with this after I've been a bad maintainer :(

davetron5000:
Note that I don't think JRuby 1.8 mode will ever pass due to:

```
Gem::InstallError: i18n requires Ruby version >= 1.9.3.
```

Is that not an issue on other PRs?

steveklabnik:
All of that is allow-fail, so it shouldn't matter.

I've been considering just removing it from CI, but haven't yet.

davetron5000:
OH, I see.  OK, this seems to be working now.

Also, I learned that `master` is not the main branch, which makes me think I might could open my other PR again.  It explaines why the rebase I was trying to do was not working.

steveklabnik:
Thank you!

Issue #1172, opened 2014-01-29
resque-web removes not the failed jobs I ask it to


x-yuri:
First, I realized that numbering scheme for jobs are somewhat odd. I've got 34 failed jobs now, and the numbers go (judging from "Remove" links): page 1 (`-7..-20`), page 0 (`19..0`). And indeed, clicking "Remove" on job `-14` removed job `-19`. The same supposedly goes for retrying jobs. I'm running `resque-1.25.1`.

lukeasrodgers:
This particular incarnation of the "resque web failure UI mapping to the wrong ID" should be fixed on 1-x-stable, though caveats about odd numbering scheme still apply.

PR #1206, opened 2014-04-17
fix Resque::Failure::Redis#each id assign when reversed order

Adrian2112:
`resque-web` on the failure tab the job ids in the queue where miscalculated and the jobs where not correctly paged 
`Resque v1.25.2`

This fix issue #1172

### DIFF
--- a/lib/resque/failure/redis.rb
+++ b/lib/resque/failure/redis.rb
@@ -39,16 +39,26 @@ module Resque
       end
 
       def self.each(offset = 0, limit = self.count, queue = :failed, class_name = nil, order = 'desc')
-        all_items = Array(all(offset, limit, queue))
-        reversed = false
-        if order.eql? 'desc'
-          all_items.reverse!
-          reversed = true
+
+        reversed = order.eql? 'desc'
+
+        # calculate offset and limit to get the reverses elements
+        if reversed
+          queue_size = self.count
+          reverse_offset = [queue_size - offset - limit, 0].max
+
+          limit = [queue_size - offset, limit].min
+          offset = reverse_offset
         end
+
+        all_items = Array(all(offset, limit, queue))
+
+        all_items.reverse!  if reversed
+
         all_items.each_with_index do |item, i|
           if !class_name || (item['payload'] && item['payload']['class'] == class_name)
             if reversed
-              id = (all_items.length - 1) - (offset + i)
+              id = offset + limit - i - 1
             else
               id = offset + i
             end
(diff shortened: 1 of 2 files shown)


steveklabnik:

Thanks! / sorry :frowning:

lukeasrodgers:
@steveklabnik I had been talking with @yaauie about addressing the issue, but if others are willing to provide a fix, I'd be happy to test it out. As mentioned above, I believe there are a couple issues with this PR. I think the ideal fix would address both the negative/incorrect failure indexing, and also the counter-intuitive failure ordering, introduced here: https://github.com/resque/resque/commit/10912efdcec91de7486e273f203c884db5f523f8

steveklabnik:
I would still like to get everyone back on master for the next release. I've been merging some stuff related to failure counts, maybe this is okay now? Anyway, this still needs a rebase too, maybe doing so will let us evaluate the failures properly.

lukeasrodgers:
@steveklabnik I will try out resque master with our app, and see if the failure count issues are resolved/improved, probably won't get around to it till next week though. Thanks for your efforts on this.

steveklabnik:
Thanks! A small note: you should try `1-x-stable`, not `master`. I haven't switched it over yet, iirc.

lukeasrodgers:
@steveklabnik I tried out 1-x-stable, the UI now appears to correctly map failures to their IDs, so far, though there appears to be a slightly related regression, as mentioned here: https://github.com/resque/resque/pull/1374.

Also, the pagination UI is still highly counterintuitive, as per above. See attached for an example. 
screen shot 2016-01-19 at 12 31 48 pm

Additionally, I think changing the sort order (currently not possible via the UI, but still semi-supported by adding `order=asc` to the querystring) is counterintuitive. If I change from the default "desc" to "asc", each page is reordered, but the overall ordering is unchanged (i.e. page 0 is still farther to the right, and lists the same set of failures, 0-19). 

I would be happy to work on this, though probably wouldn't be able to complete it for a couple weeks due to work/family commitments.

steveklabnik:
Adding to the milestone so we can make sure to track the regression.

steveklabnik:
Okay, this regression is the final blocker for 1.26.0. Gotta figure it out.

steveklabnik:
I have submitted https://github.com/resque/resque/pull/1427 to fix the regression. 

In the meantime, I _think_ that this PR's issue was fixed with other means, so let's track that issue there, and close this  PR. Thanks all!

PR #1207, opened 2014-04-24
Worker heartbeat-based cleanup

sirupsen:
Commits need to be cleaned up, I'll do that later if this has interest upstream.

At Shopify, we run Resque workers in Docker containers across a lot of different servers. Each container, and therefore worker, thus has its own hostname. If a worker is shutdown abnormally, and doesn't get to unregister itself, there's no longer anything to clean up after it because the old logic works on a per host basis.

This takes a different approach, where each worker sends heartbeats at a specified interval. When new workers start, instead of looking for dead workers on the same host—it looks for workers which haven't sent a heartbeat for a while and unregisters them.

We've run this in production on 100s of workers restarting many times per day during deploys, and it's been working great.

Note that this is a PR to the 1-x-stable branch.

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -526,7 +526,6 @@ module Resque
     def prune_dead_workers
       heartbeats = redis.hgetall(WORKER_HEARTBEAT_KEY)
       all_workers = Worker.all
-      known_workers = worker_pids unless all_workers.empty?
 
       all_workers.each do |worker|
         id = worker.to_s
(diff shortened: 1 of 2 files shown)


dylanahsmith:
@steveklabnik any interest in the ability to prune remote workers like this upstream? We rebased this on top of v1.25.2 in commit https://github.com/Shopify/resque/commit/cc245ca61a9621e2a6f46defdc2095d3f8ca5f3b.

cc @fw42

steveklabnik:
I would be interested yes, especially since you all have been using this in production.

My interest in adding stuff to Resque is directly correlated with "how sure can I be that this won't break everyone", and so production usage is a pretty good indication that it won't do that.

Sirupsen:
Yeah, we've been using this for two years soon.

steveklabnik:
Excellent.

PR #1210, opened 2014-04-27
Create a class between Resque classes and what they do in redis

davetron5000:
## Problem

I want to write code to access information about what's going on in Resque, but the code needs to work for multiple Resque instances in the same Ruby VM.  Because `Resque.redis` is global, it is very difficult (impossible in some cases) to use the Resque API directly.

## Solution

Provide an API that does not rely on a global variable that encapsulates all the ways in which Resque interacts with Redis, namely the names of keys and what sort of data structure is expected in those keys.

Consider a call like this:

```ruby
decode redis.lpop("queue:#{queue}")
```

This should mean "decode the job on queue `queue`", but it actually means "decode whatever is in redis under the key `"queue:#{queue}"` which just so happens to be how we store queues, but don't worry about that right now, just go in Redis and do it".

With this PR, it turns into this:

```ruby
decode(@data_store.pop_from_queue(queue))
```
which is saying "get me the job in queue `queue`, however that's done, and decode it.

Which means that someone _else_ can do this without knowing how to construct the redis key for queue.

And because that knowledge is now centralized in one class (`DataStore`) instead of littered throughout the codebase, one could perform these operations on multiple resque queues from the same Ruby VM, e.g. for monitoring:

```ruby
resques = {
    www: '10.0.3.4:2345',
  admin: '10.1.4.5:8765',
    ops: '10.1.4.5:8766',
}

data_stores = Hash[resques.map { |name,location|
  [name,Resque::DataStore.new(Redis.new(location))]
}]

data_stores[:www].num_failed # => how many are failed in www's Resque
data_stores[:admin].num_failed # => what about admin?
stuck_workers = data_stores[:ops].workers.select { |worker|
  data_stores[:ops].worker_start_time(worker) > 1.hour.ago
}
```
And so forth.

This is not an ideal design, but it solves the problem without breaking backwards compatibility and is better than what exists now, since it at least centralizes how Resque's data structures are stored in Redis.  It could also, in theory, allow a different backing store than Redis.

I hacked a `concerning` concept to demonstrate which calls were relevant to what—this could be split into further classes.  It's also possible that versions of the major objects (`Resque`, `Worker`, and `Job`) could be created to not use a global for `redis`, but that is for another day.

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -1,0 +1,297 @@
+module Resque
+  # An interface between Resque's persistence and the actual
+  # implementation. 
+  class DataStore
+    extend Forwardable
+
+    def initialize(redis)
+      @redis               = redis
+      @queue_access        = QueueAccess.new(@redis)
+      @failed_queue_access = FailedQueueAccess.new(@redis)
+      @workers             = Workers.new(@redis)
+      @stats_access        = StatsAccess.new(@redis)
+    end
+
+    def_delegators :@queue_access, :push_to_queue,
+                                   :pop_from_queue,
+                                   :queue_size,
+                                   :peek_in_queue,
+                                   :queue_names,
+                                   :remove_queue,
+                                   :everything_in_queue,
+                                   :remove_from_queue,
+                                   :watch_queue,
+                                   :list_range
+
+    def_delegators :@failed_queue_access, :add_failed_queue,
+                                          :remove_failed_queue,
+                                          :num_failed,
+                                          :failed_queue_names,
+                                          :push_to_failed_queue,
+                                          :clear_failed_queue,
+                                          :update_item_in_failed_queue,
+                                          :remove_from_failed_queue
+    def_delegators :@workers, :worker_ids,
+                              :workers_map,
+                              :get_worker_payload,
+                              :worker_exists?,
+                              :register_worker,
+                              :worker_started,
+                              :unregister_worker,
+                              :set_worker_payload,
+                              :worker_start_time,
+                              :worker_done_working
+
+      def_delegators :@stats_access, :clear_stat,
+                                     :decremet_stat,
+                                     :increment_stat,
+                                     :stat
+
+    # Compatibility with any non-Resque classes that were using Resque.redis as a way to access Redis
+    def method_missing(sym,*args,&block)
+      # TODO: deprecation warning?
+      @redis.send(sym,*args,&block)
+    end
+
+    # make use respond like redis
+    def respond_to?(method,include_all=false)
+      @redis.respond_to?(method,include_all)
+    end
+
+    # Get a string identifying the underlying server.
+    # Probably should be private, but was public so must stay public
+    def identifier
+      # support 1.x versions of redis-rb
+      if @redis.respond_to?(:server)
+        @redis.server
+      elsif @redis.respond_to?(:nodes) # distributed
+        @redis.nodes.map { |n| n.id }.join(', ')
+      else
+        @redis.client.id
+      end
+    end
+
+    # Force a reconnect to Redis.  
+    def reconnect
+      @redis.client.reconnect
+    end
+
+    # Returns an array of all known Resque keys in Redis. Redis' KEYS operation
+    # is O(N) for the keyspace, so be careful - this can be slow for big databases.
+    def all_resque_keys
+      @redis.keys("*").map do |key|
+        key.sub("#{redis.namespace}:", '')
+      end
+    end
+
+    class QueueAccess
+      def initialize(redis)
+        @redis = redis
+      end
+      def push_to_queue(queue,encoded_item)
+        @redis.pipelined do
+          watch_queue(queue)
+          @redis.rpush redis_key_for_queue(queue), encoded_item
+        end
+      end
+
+      # Pop whatever is on queue
+      def pop_from_queue(queue)
+        @redis.lpop(redis_key_for_queue(queue))
+      end
+
+      # Get the number of items in the queue
+      def queue_size(queue)
+        @redis.llen(redis_key_for_queue(queue)).to_i
+      end
+
+      # Examine items in the queue.
+      #
+      # NOTE: if count is 1, you will get back an object, otherwise you will
+      #       get an Array.  I'm not making this up.
+      def peek_in_queue(queue, start = 0, count = 1)
+        list_range(redis_key_for_queue(queue), start, count)
+      end
+
+      def queue_names
+        Array(@redis.smembers(:queues))
+      end
+
+      def remove_queue(queue)
+        @redis.pipelined do
+          @redis.srem(:queues, queue.to_s)
+          @redis.del(redis_key_for_queue(queue))
+        end
+      end
+
+      def everything_in_queue(queue)
+        @redis.lrange(redis_key_for_queue(queue), 0, -1)
+      end
+
+      # Remove data from the queue, if it's there, returning the number of removed elements
+      def remove_from_queue(queue,data)
+        @redis.lrem(redis_key_for_queue(queue), 0, data)
+      end
+
+      # Private: do not call
+      def watch_queue(queue)
+        @redis.sadd(:queues, queue.to_s)
+      end
+
+      # Private: do not call
+      def list_range(key, start = 0, count = 1)
+        if count == 1
+          @redis.lindex(key, start)
+        else
+          Array(@redis.lrange(key, start, start+count-1))
+        end
+      end
+
+    private
+
+      def redis_key_for_queue(queue)
+        "queue:#{queue}"
+      end
+
+    end
+
+    class FailedQueueAccess 
+      def initialize(redis)
+        @redis = redis
+      end
+
+      def add_failed_queue(failed_queue_name)
+        @redis.sadd(:failed_queues, failed_queue_name)
+      end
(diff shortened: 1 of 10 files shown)


steveklabnik:

Thanks! / sorry :frowning:

davetron5000:
Rebased.  Let me know if I can help.  This code has been running in production since I opened this PR, FWIW, as part of https://github.com/stitchfix/resque-brain

steveklabnik:
I'm going to re-queue a build here, because Travis was just being weird.

@hoffmanc what do you think of this PR?

steveklabnik:
This has a merge conflict, would you mind rebasing please? Sorry for the delay in reviewing.

steveklabnik:
(Also, I'm trying to get people back on master Resque for the next release, are there any other patches which you're using in production?)

davetron5000:
I don't know if I can.  I got some very strange conflicts and it looked there has been a ton of churn.  Let me try squashing my commits and opening up a new PR

davetron5000:
OK, I think this PR is fundamentally incompatible with the `Resque::Backend` concept.  I don't know if that solves my original problem, but I don't see how this PR can go forward in this state.

davetron5000:
I didn't realize `master` was not the main branch.  Let me take a day or two to see if I can get this rebased

steveklabnik:
Yeah, sorry. I am hoping to switch it soon, along with the next release.

steveklabnik:
Did we ever get this rebased?

chrisccerami:
I'm very :+1: this PR. Anything that makes the Ruby code we all use easier to read.

davetron5000:
OK, this is rebased.  Seeing if the build passes.  Assuming it does or I can fix any issues that come up, before merging this we should all agree that we are confident in the test coverage.  I went to great lengths to make this an internal refactor that doesn't break any interfaces, but it's obviously a lot of code to move around.

steveklabnik:
This is still causing a lot of failures, and is somehow out of date again :(

davetron5000:
rebased

steveklabnik:
Thanks!

Something funny is going on. The tests are failing, with what looks to be a legit failure on 1.8.7, and timeout errors with a LOT of extra output on the other rubies.

I also realized I hadn't weighed in on this PR yet, really, other than administrative things :smile: 

I am happy for refactorings like this, generally. I basically agree with all of your inline comments. Names are hard, and there's always more to do, but this is an improvement. :smile: 

I'm about to cut 1.26.0, so I think that, unless you know what these failures are about and can fix them right away, that I might merge this afterwards. I'm a little scared of merging a refactoring like this right before a release.

That said, I don't have many PRs in the queue, and the stuff that made this get out of date should be basically over at this point, so I don't think that that should have a "please rebase this PR 1000 times" problem, either. Is that okay? I don't plan on taking so long to release new versions in the future, so we can still get a release with this in it reasonably soon; it won't be two years if it misses 1.26.0 :smile:

davetron5000:
Yeah, you should not wait for this to do the release—we'll get it when we can and I agree you should release what you have ready right now.  I'll see about running the tests locally and figure out what's what.

steveklabnik:
Great!

Oh, one other administrative thing to make this easier: With the release of 1.26.0, I'm going to switch to `master` as the default branch. But i'm gonna keep `1-x-stable` around so that you don't need to worry about it, and then I'll merge it over myself.

davetron5000:
ok, cool.  When you do the switch will `master` and `1-x-stable` be the same, and then future works just happens on `master`?  Or will this resurrect what was on `master` previously?

steveklabnik:
Future work will happen on master, yeah. Basically, I made the old master into a `two-oh-is-cancelled` branch to save all that work, set `master` to `1-x-stable`, and force pushed.

lib/resque/data_store.rb
```diff
@@ -0,0 +1,297 @@
...
+
+    # Returns an array of all known Resque keys in Redis. Redis' KEYS operation
+    # is O(N) for the keyspace, so be careful - this can be slow for big databases.
+    def all_resque_keys
+      @redis.keys("*").map do |key|
+        key.sub("#{redis.namespace}:", '')
+      end
+    end
+
+    class QueueAccess
```
chrisccerami: Is there a reason these classes don't get their own files, since they are full fledged classes?

chrisccerami:
Aha! It seems like the timeouts are being caused because at some points in the tests, `Resque.redis =` is being passed a `Resque::DataStore` as an argument, and this case is never handled in the new definition of `Resque.redis=`. If you add

```ruby
    when Resque::DataStore
      @data_store = server
```

to the `case` statement, the timeouts should go away. It seems like it was creating a new `Redis::Namespace` with the `Resque::DataStore` set as the `redis` argument, and then chaining those many levels down.

chrisccerami:
I was able to fix the failing spec around `Redis.remove` with the following change:

```ruby
      def remove_from_failed_queue(index_in_failed_queue,failed_queue_name=nil)
        failed_queue_name ||= :failed
        hopefully_unique_value_we_can_use_to_delete_job = ""
        @redis.lset(failed_queue_name, index_in_failed_queue, hopefully_unique_value_we_can_use_to_delete_job)
        @redis.lrem(failed_queue_name, 1,                     hopefully_unique_value_we_can_use_to_delete_job)
      end
```

This was required because in `Redis.remove`, queue defaults to nil:

```ruby
      def self.remove(id, queue = nil)
        check_queue(queue)
        data_store.remove_from_failed_queue(id, queue)
      end
```

Passing an explicit `nil` argument is different from not including the argument, so `nil` was used as the `queue_name` in `remove_from_failed_queue` instead of the default of `:failed`.

chrisccerami:
I believe the failure on line 250 of `test/resque_test.rb` is because `DataStore#all_resque_keys` should be:

```ruby
    def all_resque_keys
      @redis.keys("*").map do |key|
        key.sub("#{Resque.redis.namespace}:", '')
      end
    end
```

rather than

```ruby
    def all_resque_keys
      @redis.keys("*").map do |key|
        key.sub("#{redis.namespace}:", '')
      end
    end
```

chrisccerami:
The third and final failure I'm seeing is from line 158 of `test/worker_test.rb`, where previously it read:

```ruby
Resque.redis.stubs(:get).raises(Redis::CannotConnectError)
```

it should now read:

```ruby
Resque.data_store.redis.stubs(:get).raises(Redis::CannotConnectError)
```

or the equivalent:

```ruby
Resque.redis.redis.stubs(:get).raises(Redis::CannotConnectError)
```

although I find the first one clearer.

chrisccerami:
@davetron5000 I can either push up a commit with the changes mentioned above, or you can feel free to steal them and get this PR good to go on your own. Just let me know what's easiest for you.

davetron5000:
@chrisccerami thanks so much for your comments and time!  Sorry I was on vacation or woud've responded sooner.  If you have those improvements handy, push them up and I'll steal from you :)

Also, there's no reason for the classes being in one file other than that they were experimental and I wanted to get some feedback before doing too much.  I can/will break them out into their own files.

chrisccerami:
@davetron5000 I've opened a PR to this PR 😜

https://github.com/stitchfix/resque/pull/1

chrisccerami:
@davetron5000 if you could merge https://github.com/stitchfix/resque/pull/1 into your branch and then rebase, I think this would be ok to merge.

davetron5000:
I think I've pushed the rebased branch, but GH is having issues so maybe the web UI isn't showing it?

chrisccerami:
Yeah, Travis also didn't pick up the changes, likely due to Githubs issues.

davetron5000:
ok, now looking at this I think I screwed this up.  Let me try again

davetron5000:
Hmmm.  I'm really confused now.  Here's what I did:

```
> git checkout resque-redis-interface
> git reset --hard bcdaa1c
> git fetch resque # which is the canonical repo
> git rebase resque/1-x-stable
> # fix a conflict in multiple.rb
> git push --force origin resque-redis-interface
```

Not sure what happened.  Any ideas?  Sorry I usually don't mess up with git like this :(

chrisccerami:
@davetron5000 it seems like what happened is you somehow pulled in the changes from `master` that weren't on `1-x-stable` yet. There's a few solution I can see.

1. You could clean up the commits manually that don't belong in this PR.
2. We could merge it into `1-x-stable` since those other changes have already been merged into `master` anyway, and we're going to cherry-pick your stuff to the right place.
3. You could reopen the pull request and point it to `master` with the proper commits in it.

I feel a little uneasy about 2, but perhaps @steveklabnik has an opinion.

davetron5000:
@chrisccerami OK, I think I fixed it.  Your confirmation that I pulled in stuff from the wrong branch was helpful.  I did a `git reset --hard resque/1-x-stable`, then cherry-picked my comment, then yours, and now hopefully it's clean :)

chrisccerami:
This looks good to me. I'm going to merge this PR into `1-x-stable` now, and then I'll cherry-pick it into `master`. I believe we're cutting a new release in a couple of days, so I may wait to move it into master until after that point, and this will be in the following release.

davetron5000:
Awesome, thank you for getting this through!!

Dave

fw42:
This looks very helpful to us (at Shopify) and has been bugging me for a while. Thanks for this contribution!

Issue #512, opened 2012-02-13
before_dequeue seems to not work properly


imarzi:
The before_dequeue hook doesn't seems to work even when forced on false.
I've tried and re-tried, nothing, is this an issue or a bug of mine?

Here's the entire, simple, code:
1. Job

``` ruby
require 'resque'
require 'logger'


class LoggedJob

  def self.before_perform_log_job(*args)
    log = Logger.new('test.log')
    log.info "About to perform #{self} with #{args.inspect}"
  end
end

class Job < LoggedJob

  @queue = :default

  def self.before_enqueue_test(*args)
    log = Logger.new('test.log')
    log.info "Before Enqueue"
  end

  def self.before_dequeue_test(*args)
    false
  end

  def self.perform(*args)
    rx = %x[ifconfig eth0].split(' ')[34].split(':')[1]
    tx = %x[ifconfig eth0].split(' ')[38].split(':')[1]
    puts "Params --> #{args.inspect} Sending your usage #{rx} : #{tx}!"
  end
end
```
1. App

``` ruby
require 'rubygems'
require 'resque'
require 'job'

servers = Array.new
servers << '127.0.0.1'
servers.each do |server|
  Resque.enqueue(Job, server)
end

```

PR #1227, opened 2014-05-23
Update HOOKS.md to clarify before_dequeue usage

davetapley:
This one just caught me out, and seemingly the owner of #512 as well, thus I propose it should be clarified in the docs.


maxp-hover:
I managed to miss this, and after a couple hours debugging ended up here 🤦

PR #1242, opened 2014-07-25
Make Failure::Multiple work with Failure::RedisMultiQueue back port to 1.x

tneems:
This is a back port of #1241


steveklabnik:
Thanks!

PR #1275, opened 2015-01-08
Don't reconnect to redis when we aren't forking for jobs.

dylanahsmith:
## Problem

When running with `ENV["FORK_PER_JOB"] = 'false'` resque 1.x will reconnect to redis after popping a job from the redis queue.  This doesn't make sense, because we already have a valid connection to redis, since that is how it reserved the job.  This only makes sense after forking.

## Solution

Make reconnect conditional with `if will_fork?`

Most of the code changes are to test reconnecting to redis with forking rather than without forking, where I used a pipe to collect information from the child process.

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -206,7 +206,7 @@ module Resque
             unregister_signal_handlers if will_fork? && term_child
             begin
 
-              reconnect
+              reconnect if will_fork?
               perform(job, &block)
 
             rescue Exception => exception
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -213,6 +213,21 @@ ensure
   $stderr = orig_stderr
 end
 
+def capture_io_with_pipe
+  orig_stdout, orig_stderr = $stdout, $stderr
+  stdout_rd, $stdout = IO.pipe
+  stderr_rd, $stderr = IO.pipe
+
+  yield
+
+  $stdout.close
+  $stderr.close
+  return stdout_rd.read, stderr_rd.read
+ensure
+  $stdout = orig_stdout
+  $stderr = orig_stderr
+end
+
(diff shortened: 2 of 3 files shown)


dylanahsmith:
I locked the i18n gem to version 0.6.11 to fix a travis error for ruby 1.8.7 and 1.9.2

```
Gem::InstallError: i18n requires Ruby version >= 1.9.3.
An error occurred while installing i18n (0.7.0), and Bundler cannot continue.
Make sure that `gem install i18n -v '0.7.0'` succeeds before bundling.
```

This was unrelated to this pull request though and can be reproduced on origin/1-x-stable with a `bundle install`.  Should I make a separate pull request for that commit to lock the i18n version for travis?

steveklabnik:

Thanks! / sorry :frowning:

steveklabnik:
I'm going to give this a close, then. Sorry again :(

dylanahsmith:
Sorry, I tried to rebase this earlier but I think the test suite was failing on master at the time, then just forgot about it.

I'll try again today, since this is already a part of our shopify fork.

steveklabnik:
Ah no worries!

> I'll try again today, since this is already a part of our shopify fork.

I would be interested in doing work to get shopify back on the 'real' 1.x, if that's something you all are interested in.

dylanahsmith:
This PR has been rebased now.

> I would be interested in doing work to get shopify back on the 'real' 1.x, if that's something you all are interested in.

Great!  I've managed to get @fw42 to start working on it.

steveklabnik:
Awesome, thanks! Basically, I'm all focused on 1-x now, 2.0 is cancelled.

Maybe if we can figure out how much effort it will be, we can make that the next release. I was thinking about trying to put another out, it's been a while.

PR #1276, opened 2015-01-17
Remove methods which refer to non-existent object

ice799:
Hi:

On the 1-x-stable branch, there is no Config object. The removed code ends up causing a collision with Ruby's (now deprecated) Config and produces a confusing backtrace if the `self.config` method is used.

Since the Config object doesn't exist in this branch, perhaps these methods should be removed?

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -51,18 +51,6 @@ module Resque
 
   extend ::Forwardable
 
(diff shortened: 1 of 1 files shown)


steveklabnik:
Thanks so much. This was due to me being bad at refactoring :(

PR #1278, opened 2015-01-22
Add license to gemspec

chrismar035:
I was trying to determine the licenses for my dependencies. I found that the RubyGems API will return the license, but only if it is specified in your gemspec file.


mcfiredrill:
Thank you!

Issue #1270, opened 2014-12-02
ArgumentError in RedisMultiQueue in 1-x-stable


jonah-williams:
Commit 10912efdcec91de7486e273f203c884db5f523f8 introduced a call to [`Resque.list_range(queue, offset, limit, order)`](https://github.com/resque/resque/blame/1-x-stable/lib/resque/failure/redis_multi_queue.rb#L37). Unfortunately `list_range` is only accepts 3 arguments: [`def list_range(key, start = 0, count = 1)`](https://github.com/resque/resque/blob/1-x-stable/lib/resque.rb#L247 ).

I can't find any 4 argument version of `list_range` where this would be a valid call. As a result `RedisMultiQueue` is broken in `1-x-stable` as of at least 1.25.0. Anyone attempting to view `/resque/failed` in Resque web will always hit an exception.

lisadu:
Any update on when this can be merged? `RedisMultiQueue.all` is completely broken in the current version.

guilherme:
i have my thoughts on this, check #1283

steveklabnik:
I have merged https://github.com/resque/resque/pull/1374 which will fix this.

tneems:
Keep up the good work. Glad to see movement again.

steveklabnik:
Thanks. I don't have a ton of time, but I'm gonna do my best.

jonah-williams:
Thanks @steveklabnik, nice to see an issue fixed with whatever time you can spare.

PR #1283, opened 2015-02-04
fixing multi failure queue ArgumentError , this fix: #1270

guilherme:
I've checked, actually there's a pull request for this (without commit) #1270.

First i thought that the ordering implementation must be in the Resque#list_range, since its responsability is to bring the list, ordered  would be an option. IMHO, the other methods should only rely on this behavior, and should not make any "#reverse!" and "calculate offsets" logic. If the implementation of ordering be in the #each would cause to every new failure class i would need to implement this behavior, i don't think this is ok.
And make this change have another problem: since the "ids" are the index numbers of the jobs on the queue, all the operations made with the inverse order would need some calculations to get the right id (for e.g. to make requeue and remove methods).  And i've noticed that when its filtered by class, the ids are wrong too.

I prefer to remove this 'order' stuff, and so we could think about a better solution on these all other stuff.

Currently i'm attaching the commit that removes the order, and make the 1-x-stable /failed with multi failure queue work.

### DIFF
--- a/lib/resque/failure.rb
+++ b/lib/resque/failure.rb
@@ -76,8 +76,8 @@ module Resque
     end
 
     # Iterate across all failures with the given options
-    def self.each(offset = 0, limit = self.count, queue = nil, class_name = nil, order = 'desc', &block)
-      backend.each(offset, limit, queue, class_name, order, &block)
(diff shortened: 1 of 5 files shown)


steveklabnik:

Thanks! / sorry :frowning:

steveklabnik:
I'm going to give this a close, then. Sorry again :(

PR #1288, opened 2015-02-24
Math tweaks for failures

ryanschwartz:
Fix id on paginated failures pages (resque-web.domain.tld/failed?start=1960), items would be displayed with negative ids (resque-web.domain.tld/failed/requeue/-1941), preventing retry/remove.

Fix numbering - helper would display "Showing 1960 to 1980" for items 1960-1979.

PR #1294, opened 2015-04-14
fix #1293 Resque web doen't retry the correct failed job

### LINKED ISSUE #1293: Resque web doen't retry the correct failed job on the second page of the failed jobs
Hi resque,

If I have more than 20 failed jobs in resque web and I want to retry a job on the second page. I doesn't restart the correct job! On the first page of the failed jobs in resque web the correct jobs are started.

I.e. if I retry worker with argument 27 (which is on the second page). I see in the log worker 17 is retried.

I've made a little test to confirm this problem.
Worker: I use the id to keep track of the job. I put in a `raise` to intentionally fail the job.
```ruby 
class TestWorker
  @queue = :test_queue
  
  def self.perform(id)
    Rails.logger.info "test #{id} worker ! #{Time.now.inspect}"
    puts "test #{id} worker! #{Time.now.inspect}" 
    
    raise    
  end
end
```

to add 30 workers 
```ruby 
    for i in 0..29
      Resque.enqueue(TestWorker, i)
    end
```

gems
```ruby 
gem 'resque', '1.25.2'
gem 'resque-scheduler', '2.5.5'
gem 'resque_mailer', '2.2.7'
```

ruby-1.9.3
rails 3.2

Kind Regards 
Marcel

marcelrouw:
fix #1293 Resque web doen't retry the correct failed job on the second
page of the failed jobs

### DIFF
--- a/lib/resque/failure/redis.rb
+++ b/lib/resque/failure/redis.rb
@@ -48,7 +48,7 @@ module Resque
         all_items.each_with_index do |item, i|
           if !class_name || (item['payload'] && item['payload']['class'] == class_name)
             if reversed
-              id = (all_items.length - 1) - (offset + i)
+              id = offset + (all_items.length - 1 - i)
             else
               id = offset + i
             end


greggawatt:
This is failing for Ruby 1.9.2 and 1.8.7, have you tested your changes in those versions?

marcelrouw:
sorry, I only checked with ruby 1.9.3

marcelrouw:
Hi greggawatt,

I looked at the my code and it's only a one line change with a change in a calculation. If I look at the errors of continuous integration tests. It looks like it fails on Bundler while installing i18n gem and I didn't change the Gemfile. So I don't think these error are produced by my change. 

I think the problem is cause by the "new" 0.7.0 version of i18n gem because is may not support older rubies anymore.

error:
An error occurred while installing i18n (0.7.0), and Bundler cannot continue.

Kind Regards,
Marcel

steveklabnik:
I think the PR i just merged may fix the error, restarting travis

hoffmanc:
image

There are a couple of confusing things about this travis run:

1. it didn't even run 2.x rubies
2. it states there are no environment variables set for 1.9.2, which should not be the case.

hoffmanc:
@steveklabnik #1288 is essentially a dup of this with another fix for pagination; I'd close this one out, and take #1288, after making the PR description into the respective commit messages, as they are better.  I'll request that of the OP.

steveklabnik:
@marcelrouw does that seem about right to you?

marcelrouw:
As far as I can see by just looking at the code #1288 fixed it at the same way as I did. For me it's okay the other branch is used to merge into the master of resque web.

steveklabnik:
Great, will do that. Thank you for your efforts, @marcelrouw

PR #1297, opened 2015-04-21
Adding hooks to intercept worker timeout

dbose:
While coding a gem to accelerate indexing of ElasticSearch by partitioning through resque/redis, I felt the need to trap event when resque worker timed out after configured `interval` (which defaults to 5 and can be passed as a parameter).

There is no easy way to gracefully self-terminate workers after Redis queue is empty. Of course its certainly possible to spin a `Thread` to check the Redis queue after some configured interval and invoke `worker.shutdown`. But a threaded solution does add overhead as we plan to run this across an auto-scale AWS EC2 compute nodes. `EventMachine` seems overkill.

```ruby
Resque.after_timeout do |worker|
  on_completion(worker)
end
```

Of course we can add a `after_perform` hook, but it's a JOB HOOK as opposed to worker hook, and fails to set the `@shutdown` instance variable of the `worker`. 

```ruby
def self.after_perform_check_for_empty(resource_name, from, to)
  self.on_exhausted do          
    self.shutdown rescue nil
  end
end
```

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -162,6 +162,29 @@ module Resque
     register_hook(:after_fork, block)
   end
 
+  # The `after_timeout` hook will be run in the **parent** process
+  # only once, if its times out while waiting for a job. Be careful- any
+  # changes you make will be permanent for the lifespan of the
+  # worker.
+  #
+  # Call with a block to register a hook.
+  # Call with no arguments to return all registered hooks.
+  # @overload method()
+  #   Return the existing method hooks
+  #   @return (see #hooks)
+  # @overload method(&block)
+  #   @return (see #register_hook)
+  def after_timeout(&block)
+    block ? register_hook(:after_timeout, block) : hooks(:after_timeout)
+  end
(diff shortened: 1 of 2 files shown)


hoffmanc:
:-1: no tests

hoffmanc:
@steveklabnik is there a way to auto-rebase these PRs against 1-x-stable?  I see this is failing due to the i18n gem version mismatch again...

steveklabnik:
GitHub has no "auto-rebase", no, but I don't think that's the issue here. I think it's that Travis is caching stuff.

hoffmanc:
I see - well I guess I just have to look at the builds then, and manually exclude ones that are fixed upstream.

steveklabnik:
Would you mind rebasing please to see if it fixes the tests? Sorry for the delay in reviewing.

steveklabnik:
Okay, I'm going to give this PR a close. Let me know if you're still interested in pursuing  it!

PR #1318, opened 2015-08-11
1 x stable fix ruby2.1.x specs

lukeasrodgers:
This test should never have passed, though for some reason it passes with ruby 1.9.3. I wasn't able to isolate the reason, but it had to do with silent dependencies on other tests, and the fact that test order
is not randomized (if you remove all other tests from this file, and run it on 1.9.3, this test will fail).

Tests now pass on 1.9.3, 2.0.x, and 2.1.x, as far as I can tell.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 Gemfile.lock
 test/dump.rdb
 test/dump-cluster.rdb
+test/stdout
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,19 @@
 rvm:
-  - 1.8.7
-  - 1.9.2
   - 1.9.3
   - jruby-18mode
   - jruby-19mode
   - rbx-18mode
   - rbx-19mode
+  - 2.0
+  - 2.1
 matrix:
   allow_failures:
     - rvm: jruby-18mode
     - rvm: jruby-19mode
     - rvm: rbx-18mode
     - rvm: rbx-19mode
+  include:
+    - rvm: 1.8.7
+      gemfile: Gemfiles/Gemfile.1.8.7
+    - rvm: 1.9.2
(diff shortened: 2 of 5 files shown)


yaauie:
@lukeasrodgers could you add the other versions of ruby to the travis build matrix? I'd love to validate your fixes manually on each ruby, but travis'll be better at making sure no one breaks a ruby 2.x build :smile: 

Also, the build fails on Rubies <= 1.9.2, largely because of dependencies that no longer support those rubies; we may want to add special Gemfile.lock files for travis to use for these older rubies that pins us to supported dependencies.

lukeasrodgers:
@yaauie sure I'll look into this.

lukeasrodgers:
I'll squash these commits if/when I finally get a green build.

lukeasrodgers:
@yaauie build is green. I would normally put all dependencies in the gemspec, and maybe conditionally require different versions by testing `RUBY_VERSION`, but not sure if that is desirable in this scenario (more complex than I'm used to dealing with).

Also, because tests consistently fail with 1.8.7 and redis >= 3.0.7, it might be worth moving that requirement into the gemspec itself, on the assumption that it represents a real issue with the gem rather than just an issue with the tests. E.g.

```ruby
if RUBY_VERSION == 1.8.7
  s.add_dependency "redis", "~> 3.0", "< 3.0.7"
end
```

Let me know what you think when you have a chance.

yaauie:
The trouble with moving that into the gemspec itself, is that the gemspec is evaluated at build-time, which would require us to build and distribute a separate gem for 1.8.7, something I'm not entirely willing to take on. 

I think this looks like a great approach; I plan to check on a couple things before merging tomorrow. Thanks :+1:

lukeasrodgers:
Sounds good -- I squashed those commits.

PR #1322, opened 2015-08-13
1-x-stable ruby 2.2 compatibility: add test dependency on test-unit.

### LINKED ISSUE #1295: Resque broken with ruby 2.2
Attempting to run resque under ruby 2.2 (rails 3-2-stable from github) fails silently with no output.

```
root@986255b26a04# ruby --version
ruby 2.2.0p0 (2014-12-25 revision 49005) [x86_64-linux-gnu]
root@986255b26a04# bundle --version
Bundler version 1.7.12
root@986255b26a04# bundle exec rails --version
Rails 3.2.21
root@986255b26a04# bundle exec rake resque:work RAILS_ENV="development" QUEUE=* VERBOSE="1" BACKGROUND="no" --trace; echo $?
0
```

The command returns immediately with no output. I notice a daemonized resque process that runs for a few seconds and then exits. No information is available on what happened.

Any ideas?

lukeasrodgers:
Ruby 2.2 moves this library out of stdlib into a gem.

This may fix #1295, though I was unable to produce that particular failure.

Running the command in question there, with a slight modification (adding `envrionment`), is currently working fine for me with this branch and ruby 2.2.2:

```
bundle exec rake environment resque:work RAILS_ENV="development" QUEUE=* VERBOSE="1" BACKGROUND="no" --trace; echo $?
```

Also @yaauie: all tests seem to pass fine with test-unit ~> 2.0 as well, so we could relax that dependency if you think it would make sense.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ rvm:
   - rbx-19mode
   - 2.0
   - 2.1
+  - 2.2
 matrix:
   allow_failures:
     - rvm: jruby-18mode
--- a/Gemfile
+++ b/Gemfile
@@ -11,4 +11,5 @@ group :test do
   gem "i18n"
   gem "minitest", "4.7.0"
   gem "json"
+  gem "test-unit", "~> 3.0"
 end


yaauie:
kicking the build due to a flapping test :confounded:

yaauie:
I think bumping up a test dependency is perfectly fine within semver, since it's not part of the distributed package. :+1:

Issue #1323, opened 2015-08-13
Use LPUSH to add job to front of queue


thenovices:
I would love to have the ability to add a job to the front of the queue. In particular, I currently want to do this in conjunction with resque-retry: in certain cases, I want to have the job immediately re-run after an Exception, but without having to go to the end of the queue.

Here is a Resque plugin that does this by using LPUSH instead of RPUSH on the queues:
https://github.com/coupa/resque-priority_enqueue

It's not really maintained though, and I would love to see it included in the main Resque project. It's only a couple lines of code.

yaauie:
If you'd like to create a pull-request, I'll consider it against 1-x-stable as long as it comes with good test coverage; since the project is also MIT-licensed, there won't be a problem with copying code from it wholesale.

Jacobkg:
@steveklabnik Looks like there was a PR for this feature that was closed https://github.com/resque/resque/pull/1324#issuecomment-155817070. Should we retag this as 'feature request'?

steveklabnik:
:+1:

chrisccerami:
It sounds to me like #1324 was closed with the idea that Resque isn't interested in a feature to use LPUSH because you can just add jobs to a higher priority queue. If that's the case, then should this issue also be closed?

fw42:
Has this been addressed by https://github.com/resque/resque/pull/1425? Can this issue be closed?

chrisccerami:
Seems like it to me.

nathantsoi:
@fw42 & @steveklabnik, unless i'm missing something here, lpush is necessary to implement a LiFo queue and cannot be implemented w/ a "higher priority queue." If I submit a pr against the latest branch, would the Resque team consider merging again? As noted in https://github.com/resque/resque/pull/1425, this feature has been lost in the latest release.

thenovices:
@nathantsoi Not sure what the concern is.
1) They had previously mentioned using a higher priority queue as an imperfect solution without having to patch resque with a lpush option. Now that we have a lpush option via #1425 no higher priority queue is needed.

2) I'm not sure why another merge is needed? #1425 got merged in, just after the latest release was cut (and this was done on purpose). It'll be included when the next release is cut.

nathantsoi:
you're totally right @thenovices, i looked in the release gem, not on master. seems to be there and all is well! looking forward to the next release :)

scomma:
[...]

PR #1324, opened 2015-08-14
Add Resque.priority_enqueue

thenovices:
Adds the ability to add jobs to the front of a queue via `Resque.priority_enqueue`
and `Resque.priority_enqueue_to`. This is done by calling redis's `LPUSH` command
instead of `RPUSH`.

Addresses #1323 

Sadly due to the use of `*args` in the relevant functions, I couldn't see an easy way to extend the existing functions, so I created duplicates. If you know of a way to DRY things up, please let me know.

Also, if you have a recommendation for better method names, let me know. I realize that priority_enqueue could potentially be confused with priority queues, which should allow assignment of a specific priority level, whereas in this case, the implementation only allows you to put a job at the front of the queue. An alternative is to use the term `unshift` instead of `enqueue`.

I added a couple of tests; let me know if there are any other tests that would be helpful.

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -217,6 +217,14 @@ module Resque
     end
   end
 
+  # Like `#push` but uses LPUSH to place the item on the front of the queue.
+  def lpush(queue, item)
+    redis.pipelined do
+      watch_queue(queue)
+      redis.lpush "queue:#{queue}", encode(item)
+    end
+  end
+
   # Pops a job off a queue. Queue name should be a string.
   #
   # Returns a Ruby object.
@@ -298,6 +306,11 @@ module Resque
     enqueue_to(queue_from_class(klass), klass, *args)
   end
 
+  # Just like `enqueue` but puts the job at the front of the queue.
+  def priority_enqueue(klass, *args)
+    priority_enqueue_to(queue_from_class(klass), klass, *args)
+  end
+
   # Just like `enqueue` but allows you to specify the queue you want to
   # use. Runs hooks.
   #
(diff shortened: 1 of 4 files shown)


hoffmanc:
Gut reaction is that you are losing all of the edge case tests by copy/pasting all that method logic, just to get lpush.

thenovices:
I don't disagree at all. I don't like having all this copy and pasting either. But it seemed to me that alternate solutions would involve significant internal API changes... Do you have any ideas about this?

hoffmanc:
Why not have a higher-priority queue, and spin up a worker with that queue listed first in `QUEUES`?

thenovices:
Yeah, that could work. The problem is that I would need a "priority" version of every single normal queue, e.g., I have some machines that only handle jobs from one queue, and so I need a priority version of that queue; having a global higher-priority queue doesn't work because then the job might go to another worker.

hoffmanc:
AFAIK, queues are cheap, so that doesn't seem like a major headache to me.  In the last company I worked for, we had at least 50 different queues, and each one had it's own high (and low) variations.  If you are following a convention (e.g., all high priority queues are the original queue name plus `_high`), the changes to whatever code is pushing jobs onto queues  and spinning up workers should require minimal code change.  But perhaps I am mistaken.

hoffmanc:
Of course, having said that, it could be useful to minimize the number of queues for management / reporting.  So, maybe worth adding a `position` parameter (or whatever you decide) and seeing what that looks like from an API standpoint?

hoffmanc:
@steveklabnik  :-1: I don't see what this gets us, that isn't already taken care of by having workers processing more than one queue.

steveklabnik:
@thenovices , I'm going to agree with @hoffmanc here. We're not going to add a ton of new features to Resque at the moment, and in the meantime, workers on more than one queue should solve this problem.

I'd prefer to focus on improving the quality of the code before we add new features, especially once which have a workaround currently. Thank you!

thenovices:
:+1: make sense, thanks!

PR #1325, opened 2015-08-18
clean up rake test output

hoffmanc:
I have begun familiarizing myself with the 1-x-stable branch, and did some cleanup after seeing a lot of warnings in the test output.


hoffmanc:
just realized I broke the rake task...

PR #1334, opened 2015-10-01
Fix backport of before_pause & after_pause hooks

stevenschobert:
Heya awesome folks! Thank you for all awesome work that goes into resque :tada: 

I ran into a small issue today, in the `1-x-stable` version (`v1.25.2`), where `before_pause` and `after_pause` hooks would never get called. Workers are still paused correctly, they just don't fire their associated hooks.

After a bit of digging, I think there was just a small mixup in #831 when the hooks got backported to `1-x-stable` (which is awesome by the way, :heart: for the backport).

I've added in the `run_hook` calls to the `pause_processing` and `unpause_processing` methods respectively, and added specs to cover the change.

Cheers! Please let me know if there's anything else I can do!

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -168,8 +168,10 @@ module Resque
     block ? register_hook(:before_pause, block) : hooks(:before_pause)
   end
 
-  # Set the after_pause proc.
-  attr_writer :before_pause
+  # Register a before_pause proc.
+  def before_pause=(block)
+    register_hook(:before_pause, block)
+  end
 
   # The `after_pause` hook will be run in the parent process after the
   # worker has paused (via SIGCONT).
@@ -177,8 +179,10 @@ module Resque
     block ? register_hook(:after_pause, block) : hooks(:after_pause)
   end
 
-  # Set the after_continue proc.
-  attr_writer :after_pause
+  # Register an after_pause proc.
+  def after_pause=(block)
+    register_hook(:after_pause, block)
+  end
 
   def to_s
     "Resque Client connected to #{redis_id}"
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -507,6 +507,7 @@ module Resque
     # currently running one).
     def pause_processing
       log "USR2 received; pausing job processing"
+      run_hook :before_pause, self
       @paused = true
     end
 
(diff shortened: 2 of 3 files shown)


courtsimas:
Yep, I +1 this.

hoffmanc:

@steveklabnik and I are going to be working on Resque again, but this PR needs some attention. If you get the chance, would you mind

making the tests pass, if you're up for it.

If you're not up for it, that's cool, just let us know so we can investigate
Thanks! / sorry :frowning:

stevenschobert:
No worries on the wait, I totally understand!

I'm totally happy to help investigate the specs failing. It likes like the specs are all passing except on ruby `1.9.2`, with the following error: 

```
  1) Failure:
#<Class:0x000000017d9438>#test_SIGTERM_and_cleanup_does_not_occur_in_allotted_time [/home/travis/build/resque/resque/test/child_killing_test.rb:25]:
<nil> expected to not be nil.

  2) Failure:
#<Class:0x000000017d9438>#test_SIGTERM_and_cleanup_occurs_in_allotted_time [/home/travis/build/resque/resque/test/child_killing_test.rb:25]:
<nil> expected to not be nil.

  3) Failure:
#<Class:0x000000017d9438>#test_old_signal_handling_just_kills_off_the_child [/home/travis/build/resque/resque/test/child_killing_test.rb:25]:
<nil> expected to not be nil.
```

After digging through some other pull-requests, I noticed that #1343 also failed for the sam reason. Any ideas as to what the root of this issue might be? Could something in this changeset have introduced that error?

hoffmanc:
Hi @stevenschobert; good question.  I closed out that PR because I didn't realize that there were separate gemfiles for 1.8.7 and 1.9.2.  Once I get that cleared up I'll rerun and see what the output looks like.

hoffmanc:
Looks like these failures are not deterministic... which makes sense, since they rely on forking another process.  I'm unable to repro locally right now, but will keep trying.

hoffmanc:
@steveklabnik :+1:

steveklabnik:
Thanks so much for this!

stevenschobert:
@steveklabnik @hoffmanc Thank you both! Happy I was able to help!

PR #1344, opened 2015-11-05
freeze airbrake for 1.8.7 compatibility

hoffmanc:


hoffmanc:
Seems useful to keep the 1.8.7 lock file around, but I can remove it if others don't approve.

hoffmanc:
@steveklabnik I think if we merge this the hash rocket - related PR test failures should go away.

As to the 1.9.2 related failures, I'm still investigating.  I'm unable to repro with 1.9.2 locally, even when I set the timeout for the blpop call to 0.

steveklabnik:
Let's check it out.

PR #1346, opened 2015-11-09
Remove repeated encode and decode method implementations

RochesterinNYC:
The encode and decode method implementations are repeated (and exactly the same) within the Resque module, Helpers module, Worker class, and Job class. Much easier to maintain with one true source of implementation (in the top-level Resque module).

Also, encode and decode wrapper methods left in the Helpers module, Worker class, and Job class for backwards compatibility (in case anyone's explicitly utilizing these methods off instances of the classes or the modules).


hoffmanc:
:+1:

steveklabnik:
Solid refactoring. Thank you!

PR #1350, opened 2015-11-18
Remove repeated constantize and classify method implementations

RochesterinNYC:
The constantize and classify method implementations are repeated. This PR moves the one true implementation of these to the Resque module and retains wrapper methods in the Helpers module and Job class for backwards compatibility.


steveklabnik:
Thank you!

PR #1351, opened 2015-11-19
Only allow before_first_fork hook to run once in worker parent

RochesterinNYC:
Implements the behavior for a TODO in the worker test.

It doesn't seem that currently we actually force the parent worker to call the `before_first_fork` hook only once. This hook is executed during the `startup` method which is called each time `worker.work` is called. If `work` is called multiple times on a worker, `the before_first_fork` hook is ran multiple times.

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -149,6 +149,7 @@ module Resque
       @queues = queues.map { |queue| queue.to_s.strip }
       @shutdown = nil
       @paused = nil
+      @before_first_fork_hook_ran = false
       validate_queues
     end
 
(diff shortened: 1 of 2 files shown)


RochesterinNYC:
@steveklabnik any thoughts on this PR or any changes you think should be made?

steveklabnik:
I haven't read this code in a while, and haven't had the time to dig into it.


@hoffmanc , what do you think?

RochesterinNYC:
Anything I can do help move this PR along?

steveklabnik:
Nope, just needed some time. I think this should be fine, sorry about the wait.

PR #1358, opened 2015-12-04
Fix NoMethodError on `working` server view

rjwalsh88:
In our application, we discovered a race in the `working` view.
Worker#job is called twice, which means on its second invocation
it could actually be nil, or have a different payload.  This fix
protects against a race during page rendering, and makes nested
hash lookups safer.

Thanks for great work on an amazing product!


steveklabnik:
Nice catch! Thanks so much.

PR #1362, opened 2015-12-09
Recreate an updated CONTRIBUTING.md and update README accordingly

RochesterinNYC:
Moving back to 1-x-stable as the canonical default branch for now means that there's no CONTRIBUTING.md anymore as it was added in master. It would be nice for potential/current contributors to have access to this in 1-x-stable.


steveklabnik:
Thank you!

PR #1363, opened 2015-12-18
Fixed 'frozen string' error when attempting to remove dead worker

dodgymike:
Fixed an issue where resque remove <worker id> caused
'<gem path>/lib/resque/worker_registry.rb:72:in `force_encoding': can't modify frozen String (RuntimeError)'

### DIFF
--- a/lib/resque/worker_registry.rb
+++ b/lib/resque/worker_registry.rb
@@ -69,7 +69,7 @@ module Resque
     # @return [Resque::Worker]
     def self.find(worker_id)
       if exists?(worker_id)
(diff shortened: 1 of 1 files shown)


steveklabnik:
Oh sorry! Master is basically gone at this point, the current situation is:

* `1-x-stable` is the current release
* `master` and `two-oh-is-cancelled` are the old 2.0 code, that's no longer happening.

I will be switching `master` to be `1-x-stable` in the near future.

As such, master doesn't get any more PRs, so while I appreciate this, it's not going to happen. Thanks!

PR #1364, opened 2015-12-18
Ensure Resque travis builds have latest version of bundler

RochesterinNYC:
- Prevents following error: `NoMethodError: undefined method 'spec' for nil:NilClass`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,9 @@ rvm:
   - 2.0
   - 2.1
   - 2.2
+before_install:
+  - gem update --system
+  - gem update bundler
 matrix:
   allow_failures:
     - rvm: jruby-18mode
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ group :test do
   gem "rake"
   gem "rack-test", "~> 0.5"
   gem "mocha", :require => false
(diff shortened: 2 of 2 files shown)


RochesterinNYC:
Saw the error (`NoMethodError: undefined method 'spec' for nil:NilClass`) in one of the builds for #1363.

Refer to PR/issue here: thoughtbot/factory_girl#846

RochesterinNYC:
Hm, Airbrake is breaking in the builds for later versions of Ruby (`>= 1.9.3`)

RochesterinNYC:
@steveklabnik Seems like Airbrake migrated from version 4 to 5 for latest release and this is the cause? Especially seeing as how the Gemfiles for `1.8.7` and `1.9.2` are locked at `4.3.2` and `4.3.3.` respectively.

https://github.com/airbrake/airbrake/blob/master/docs/Migration_guide_from_v4_to_v5.md#general-changes

I can open another PR investigating how to migrate appropriately if needed. We can lock the version in at `4.3.4` in the meantime if we want to amend these errors for builds.

steveklabnik:
Yeah, this seems ifne. I think that locking them in this way makes sense.

PR #1369, opened 2016-01-05
warns about environment load, and avoid raising an error

josedonizetti:
For a new comer to the project, getting an exception and a huge stack trace because he forgot to call the environment task might be scary. This handles the failure more gracefully.


steveklabnik:
Oh sorry! Master is basically gone at this point, the current situation is:

* `1-x-stable` is the current release
* `master` and `two-oh-is-cancelled` are the old 2.0 code, that's no longer happening.

I will be switching `master` to be `1-x-stable` in the near future.

As such, master doesn't get any more PRs, so while I appreciate this, it's not going to happen. Thanks!

PR #1374, opened 2016-01-12
Issue 1230/multi failuire backend regression

steveklabnik:


steveklabnik:
I don't understand what's up with this failure, I'm retrying now.

steveklabnik:
Travis is being extremely weird. One good build works, one other one fails. I'm going to merge.

lukeasrodgers:
This may have introduced a regression: https://github.com/resque/resque/blob/5bec955e5ca6e402fca775fbb5798355c09a6908/lib/resque/failure/multiple.rb#L59 takes one argument, but we're passing two now (`backend.remove(id, queue)`).

This causes an exception when I try to remove a failure using the web UI, running 1-x-stable
screen shot 2016-01-19 at 12 21 53 pm

steveklabnik:
... i was just about to say "let's open an issue", but I see you also opened a PR! so let's talk about it there

steveklabnik:
Oh wait, that's a link to another open PR. Same thing, we can use that to track it.

PR #1376, opened 2016-01-13
Remove helper deprecation warnings

Jacobkg:
The helper deprecation warnings were added here: https://github.com/resque/resque/commit/30f9310a4241e8026727cd049010ee74908d295d

Since there is no longer a 2.0 deprecation plan, I propose that we remove the warnings (which currently show up when you run `rake resque:work` on `1-x-stable`


steveklabnik:
Yes, thank you!

This module is still problematic, but until we have a replacement, this is just noise.

Issue #1253, opened 2014-09-15
"rake resque:workers" is not documented


wonnage:
If you don't pass an environment variable, it just exits immediately. And it's not on the FAQ or anything.

Jacobkg:
@steveklabnik There are a couple problems here. The first is that this commit broke the `abort` error message which explained which environment variables to set: https://github.com/resque/resque/commit/7d0b20ee00986b749b03d4407af199039b504cc2

The second is that if you don't set `COUNT` in `rake resque:workers` then we don't enter the loop that starts threads so no other checks are started.

I can create separate pull requests for each of these

steveklabnik:
@Jacobkg that'd be great! I just merged the first one.

steveklabnik:
With these two PRs merged, I believe this is now fixed :confetti_ball:

PR #1377, opened 2016-01-13
Fix error message for rake resque:work when QUEUE is not set

Jacobkg:
Fix 1 of 2 for #1253

The previous error message of `set QUEUE env var, e.g. $ QUEUE=critical,high rake resque:work` was broken when some code was moved from the rake task to `worker.rb` (see https://github.com/resque/resque/commit/7d0b20ee00986b749b03d4407af199039b504cc2). This PR moves back the `begin...rescue` so that it properly traps the exception and presents the error message.

NOTE: I realize that I accidentally picked up a few whitespace deletions as well. Let me know if you want me to get rid of them before you merge

### DIFF
--- a/lib/resque/tasks.rb
+++ b/lib/resque/tasks.rb
@@ -9,7 +9,12 @@ namespace :resque do
   task :work => [ :preload, :setup ] do
     require 'resque'
 
-    worker = Resque::Worker.new
+    begin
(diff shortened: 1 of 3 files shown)


steveklabnik:
Thanks! The whitespace stuff is okay. :)

PR #1378, opened 2016-01-13
Performance fixes and metrics

fw42:
This is essentially https://github.com/Shopify/resque/pull/9 (and https://github.com/Shopify/resque/pull/10), but against upstream/1-x-stable. (Also polished a few rough edges.)

@steveklabnik @sirupsen @dylanahsmith

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -487,6 +487,45 @@ module Resque
     end
   end
 
+  # Returns a hash with queue names and sizes in a single network roundtrip
+  def queue_sizes
+    queues = self.queues
+
+    sizes = redis.pipelined {
+      queues.each { |name| redis.llen("queue:#{name}") }
+    }
+
+    Hash[queues.zip(sizes)]
+  end
+
+  # Returns a hash of queues with sample of size `sample_size` from the front
+  # of the queue and the size of the queue.
+  def sample_queues(sample_size: 1000)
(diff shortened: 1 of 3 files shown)


fw42:
Tests failed because I'm using keyword arguments. Do you want 1-x-stable to be Ruby 1.9 compatible?

Sirupsen:
For context, we needed these calls for querying Resque across datacenters where we can't afford the otherwise 10s of roundtrips. We've been running this in production for ~6 months for monitoring purposes.

steveklabnik:
> Do you want 1-x-stable to be Ruby 1.9 compatible?

We need to be 1.8 compatible, yes. If there's something super, super onerous that makes sense to drop 1.8, then I will, but kwargs aren't enough to justify it :)

steveklabnik:
This looks great! Let's get Travis green and then :shipit:

fw42:
> We need to be 1.8 compatible, yes

Just out of curiosity: Why? That's ancient :)

fw42:
Tests are passing now :green_apple:

steveklabnik:
@fw42 it is, but without knowing any kind of metrics, I can't know how many people are still relying on it, even though it's EOL :( Maybe _this_ would be a good motivation for a Resque 2.0, rather than a full re-write.

Anyway, woo! Green! Let's go!

PR #1381, opened 2016-01-14
Fail rake rescue:workers if COUNT is not set

Jacobkg:
2nd of 2 PRs to address #1253 

Added a check that the COUNT environment variable is set to a number greater than 0 with an appropriate error message.

Refactored the test file a bit in order to support testing multiple rake tasks, and also switched it to using the `describe` syntax of minitest. Finally borrowed some code from Rails to silence stderr messages so as to not pollute the test output.

### DIFF
--- a/lib/resque/tasks.rb
+++ b/lib/resque/tasks.rb
@@ -23,6 +23,10 @@ namespace :resque do
   task :workers do
     threads = []
 
+    if ENV['COUNT'].to_i < 1
+      abort "set COUNT env var, e.g. $ COUNT=2 rake resque:workers"
+    end
+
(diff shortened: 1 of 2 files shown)


steveklabnik:
It looks like `IO::NULL` wasn't around in 1.8.

Jacobkg:
Yeah, just noticed that. Getting a 1.8.7 environement up and running so I can fix.

Jacobkg:
@steveklabnik I realized that there already was an error output suppresion method I could call from `test_helper.rb` so switched to that and travis is now green

steveklabnik:
Thanks!

PR #1243, opened 2014-07-25
encoding should be done outside of pipeline block to accomodate redis

beedub:
Redis 3.0 makes a change to return a future for some commands run inside pipeline or multi.  This change allows serialization (like `to_json`) to read from redis and get a value, not a future.

I'm not sure how best to test this.  The test I included may not even be a test at all... any ideas?

PR #1384, opened 2016-01-15
encoding should be done outside of pipeline block

beedub:
(reopen of https://github.com/resque/resque/pull/1243)

Redis 3.0 makes a change to return a future for some commands run inside pipeline or multi. This change allows serialization (like to_json) to read from redis and get a value, not a future.

I'm not sure how best to test this. The test I included may not even be a test at all... any ideas?

@steveklabnik

### DIFF
--- a/test/resque_test.rb
+++ b/test/resque_test.rb
@@ -9,6 +9,13 @@ describe "Resque" do
     Resque.redis = @original_redis
   end
 
+  it "can push an item that depends on redis for encoding" do
+    Resque.redis.set("count", 1)
+    # No error should be raised
+    Resque.push(:test, JsonObject.new)
+    Resque.redis.del("count")
+  end
+
   it "can set a namespace through a url-like string" do
     assert Resque.redis
     assert_equal :resque, Resque.redis.namespace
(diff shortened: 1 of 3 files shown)


steveklabnik:
Hey @beedub !

First of all, apparently there's a merge conflict here, sorry about that!

Second, I agree that I'm not sure about the tests here... hm. In the last PR, it was still failing before and passing after, is this still true? it might be Good Enough for now.

beedub:
Fixed the merge conflict

My memory is a bit faded given how long ago it was, but from what I said in the previous PR, the test I added failed before I added my logic changes, and then passed after my logic changes (ie. encoding outside of pipeline)

test/resque_test.rb
```diff
@@ -9,6 +9,14 @@
     Resque.redis = @original_redis
   end
 
+  test "can push an item that depends on redis for encoding" do
```
steveklabnik: This is causing the travis failure. It should be `it`, not `test`, like the test right below it

beedub:
updated the test and build now passes

steveklabnik:
Great! Thanks so much.

PR #1385, opened 2016-01-15
Switch all tests to describe/it Minitest syntax

Jacobkg:
Make all test files use the same syntax. This helps the test output appear in a single place instead of split up by file.

@steveklabnik


steveklabnik:
Thank you!

Issue #1141, opened 2013-10-16
VERBOSE and VVERBOSE behavior change in 1.x stable


joshuaflanagan:
I'm trying to understand why the behavior of the VERBOSE and VVERBOSE environment variables changed, and if it was intentional.

In the past, those environment variables controlled the level of logging (some messages would not be logged unless you set VVERBOSE). Now, they just control the format of the log messages.

I understand the move away from `#verbose` and `#very_verbose` methods toward using a Logger with level. However, it isn't clear why the environment variables still interact with the deprecated methods instead of setting the `Logger.level`.

I would expect VERBOSE=1 to set `Logger.level` to `info` and VVERBOSE=1 to set `Logger.level` to `debug`. That would more closely match the previous logic.

For example, before the logging overhaul, when you set VERBOSE=1, you would see messages when jobs were picked up, but you wouldn't see the `Checking queueA` and `Sleeping for 5.0 seconds` messages. You only saw those if you set VVERBOSE=1.

Now you always see those messages, even if you just set VERBOSE=1.

Being able to vary the level of logging via environment variables was very useful. Being able to vary the format of the log messages via environment variables is not.

Is there a way to vary the log level via environment variables with the new approach?

Would you be open to a PR that restores the original environment variable behavior (using the new `Resque.Logger` of course)?

steveklabnik:
I would like to revert to previous behavior here for 1.26.0.

Jacobkg:
@steveklabnik I believe that this recent commit of yours also addresses this issue: https://github.com/resque/resque/commit/0519b60f66834bd9b76008cf1acea497f025b105

steveklabnik:
That removed the deprecation, but if the behavior changed, it should be like the old behavior.

Jacobkg:
@steveklabnik Cool, I think I know what to do. Working on a PR

steveklabnik:
https://github.com/resque/resque/pull/1386 fixes this.

PR #1386, opened 2016-01-15
Restore VERBOSE and VVERBOSE

Jacobkg:
Fix for #1141

Switches log level to `debug` if `very_verbose` is set. Switches log level to `info` if `verbose` is set.


steveklabnik:
Thank you so much!

Issue #1074, opened 2013-07-10
v1.23.1 broke SemVer for the 1.x-stable line


ches:
Filing this issue per @steveklabnik's request:

The integration of c1ddf97fe006bc4 on the 1-x-stable branch violated SemVer -- as noted in the commentary on that commit, methods like `Resque.before_fork` and `Resque.after_fork` previously returned a proc (the single registered hook that was possible) when called without an argument; they now return an Array of all the registered hooks.

I was under the impression in the past that registration of multiple hooks -- long sought-after by many -- was a major feature slated for 2.0 *because of* the backward incompatibility implication.

The revision referenced (2427c97) in the commit message of c1ddf97fe006bc4 is not actually present on 1-x-stable, the "conflict fix" commit brings in the changes so there was no actual merge. That commit appears to have been rewritten by @jonhyman on #831 -- this got rather messy I'm afraid.

See also #831 and #859.

Jacobkg:
@steveklabnik I'm kinda stuck on what the actual use case is for either the current or previous behavior. It would really help me making a decision if I knew how people were relying on the fact that the hooks return a proc. @ches I doubt you are still following this but maybe you could weigh in.

ches:
@Jacobkg: I'm not sure if this answers your question, but the hooks are basically [the basis for Resque plugins][1], so I think that doc and looking at one or two available plugins will shed light on how people are using hooks. Apologies if that's much more elementary than what you're asking, let me try to presume further what you may be looking for:

The central matter here was that previously Resque only supported *one* registered [worker hook][2] per type (one `before_fork`, one `after_fork`, etc.), so if you used multiple plugins that relied on worker hooks, they would stomp on each other—essentially the last to register wins and nothing else gets run. That's pretty much what the motivating problem of New Relic's Resque instrumentation amounted to, in fact. (#537)

Clearly that's rather limiting, but that's the way it was, so the API was that `Resque.before_fork` or `Resque.after_fork`, when called without a block argument, worked like accessors to return the *one* currently registered hook of that type: they returned a single Proc. The set of changes referred to here introduced a stack/array of hooks per type so that multiple plugins could register worker hooks, but that meant that `Resque.before_fork` and `Resque.after_fork` would then return *an array* of all the registered hooks (Procs). Hence the breaking API change. A welcome change, but one that needed a major version bump by semver.

[...]

PR #1387, opened 2016-01-20
Restore 1.23.0 behavior and return block

Jacobkg:
Fix for #1074. Return the passed block for all of the hooks.

@steveklabnik

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -539,9 +539,8 @@ module Resque
 
     @hooks ||= {}
     @hooks[name] ||= []
-
-    block = Array(block)
-    @hooks[name].concat(block)
+    @hooks[name] << block
+    return block
   end
 
   # Clear all hooks given a hook name.
--- a/test/resque_hook_test.rb
+++ b/test/resque_hook_test.rb
@@ -191,4 +191,14 @@ describe "Resque Hooks" do
       file2.delete
     end
   end
+
+  [:before_first_fork, :before_fork, :after_fork, :before_pause, :after_pause].each do |hook_name|
+    it "returns the block from #{hook_name}" do
+      result = Resque.public_send(hook_name) do
+        3 * 7
+      end
+      assert_equal result.call, 21
+    end
+  end
+
 end
--- a/test/resque_hook_test.rb
+++ b/test/resque_hook_test.rb
@@ -194,7 +194,7 @@ describe "Resque Hooks" do
 
   [:before_first_fork, :before_fork, :after_fork, :before_pause, :after_pause].each do |hook_name|
     it "returns the block from #{hook_name}" do
-      result = Resque.public_send(hook_name) do
+      result = Resque.send(hook_name) do
         3 * 7
       end
       assert_equal result.call, 21


steveklabnik:
Thanks for this!

`public_send` only existed in 1.9, it's just `send` in 1.8.

Jacobkg:
Gah! I forgot about that. Fixing

test/resque_hook_test.rb
```diff
@@ -191,4 +191,14 @@ def self.perform
       file2.delete
     end
   end
+
+  [:before_first_fork, :before_fork, :after_fork, :before_pause, :after_pause].each do |hook_name|
+    it "returns the block from #{hook_name}" do
+      result = Resque.send(hook_name) do
```
steveklabnik: :+1: , since this is just a test, no need to worry about public vs non, imho

Jacobkg:
@steveklabnik Should we be concerned whether anyone since 1.23.1 has been relying on the new behavior and that this will break their code?

steveklabnik:
@Jacobkg that is a good question. I'm not sure how to answer it.

Jacobkg:
I guess we can see if anyone complains after the new release?

steveklabnik:
It also might be worth making a second method with the new behavior, and then in the upgrade notes, suggesting that you choose the correct one?

hackhowtofaq:
...and deprecate the other with a msg?

lib/resque.rb
```diff
@@ -539,9 +539,8 @@ def register_hook(name, block)
 
     @hooks ||= {}
     @hooks[name] ||= []
-
-    block = Array(block)
-    @hooks[name].concat(block)
+    @hooks[name] << block
+    return block
```
ches: Hmm yeah, I don't think this change really makes sense, hopefully my elaboration on #1074 helps make it easier to reason why.

This change wouldn't completely solve the API incompatibility of the original change, and even if it did I don't believe that it would make semantic sense: the change in #831 fundamentally changed Resque to support a stack of registered worker hooks instead of just one, so it wouldn't make sense for e.g. `Resque.before_fork` to return a single Proc anymore—what Proc would it be, some arbitrary choice of first or last hook in the stack of potentially several?

Here's why this change also incomplete: the change is in `register_hook` so that it indirectly results in its callers now returning a Proc instead of an array, but `register_hook` isn't always called, e.g.

```ruby
def before_fork(&block)
  block ? register_hook(:before_fork, block) : hooks(:before_fork)
end
```

If I call this accessor-style without passing a block argument, the `hooks(:before_fork)` branch is called instead and I still get an array returned, which is still breaking the original method contract.

ches:
To be honest, as the person who originally raised the point of the breaking API change, now that it's been out there in the wild on the 1.x line of releases I don't think it's worth trying to retroactively smooth this over, especially if no one else ever complained.

A breaking change is unfortunate but this one most likely manifests in a very straightforward, easy-to-fix way, and by now most people affected have probably already felt the sting, updated their code and moved on. For APIs to make internally consistent semantic sense I believe the whole multiple-worker-hooks feature would need to be reverted and pushed back to 2.x, which is probably much more troublesome at this stage.

Just my 2¢.

By the way @steveklabnik I hope my tone or attitude was never something that made you doubtful about maintenance, release management, or anything of the sort!

Jacobkg:
Thank you @ches for enlightening me. On that recommendation, @steveklabnik I am going to close this PR

steveklabnik:
> By the way @steveklabnik I hope my tone or attitude was never something that made you doubtful about maintenance, release management, or anything of the sort!

No worries, it didn't. Stuff hasn't been great for a while. It happens.

PR #1390, opened 2016-01-22
Pull common test setup into main before

Jacobkg:
There were a few common pieces of test reset code sprinkled around the test files. I think it will make it easier to manage and add future tests if test authors don't need to worry about this common setup and instead it always happens before every test

@steveklabnik


steveklabnik:
Agree 100%, thanks!

PR #1395, opened 2016-01-25
Optimization to avoid unnecessary smembers calls

Jacobkg:
In many Resque implementations, including my company's, the workers are given a hard-coded list of queues and do not use glob (aka `*`) matching.

In these cases, Resque still loops over each queue that the worker is listening on (which could be dozens) and then for each of them calls `Resque.queues` which triggers a `Resque.redis.smembers` call. This means that if there are 100 workers and they are listening on 10 queues, each, then we issue 1,000 `smembers` calls instead of the 0 that we need (because we already know which queues we are listening on and don't need to glob match). This can really impact the performance of the `Resque.redis` instance.

This PR adds an optimization to avoid making the `Resque.redis` call in the case that the `queue` does not contain any glob matching characters.

NOTE: There is probably a larger refactor I could do around how the `worker` stores which queues it is listening on which would be even more efficient.

Thoughts? @steveklabnik

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -327,6 +327,7 @@ module Resque
     end
 
     def glob_match(pattern)
+      return [] unless ['*', '?', '{', '}', '[', ']'].any? {|char| pattern.include?(char) } #optimization to avoid Resque.queues call if no glob characters
       Resque.queues.select do |queue|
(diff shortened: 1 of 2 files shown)


Jacobkg:
On further reflection, I want to implement this differently

PR #1397, opened 2016-01-26
Remove ruby modes that have been broken for years

Jacobkg:
We have been ignoring Travis results for rubinius and jruby since 2013. I agree that ideally Resque would support them but in the meantime I see no reason to make our Travis builds like 4x slower and waste the generous resources Travis donates to this project.


steveklabnik:
Agreed, thanks.

PR #1403, opened 2016-02-04
Fixed failed jobs list when params[:class] specified

hasaniskandar:
This will fix issue which failed jobs does not display a complete list when `params[:class]` provided.

### DIFF
--- a/lib/resque/failure/redis.rb
+++ b/lib/resque/failure/redis.rb
@@ -39,6 +39,10 @@ module Resque
       end
 
       def self.each(offset = 0, limit = self.count, queue = :failed, class_name = nil, order = 'desc')
+        if class_name
+          original_limit = limit
+          limit = count
+        end
         all_items = limit == 1 ? [all(offset,limit,queue)] : Array(all(offset, limit, queue))
         reversed = false
         if order.eql? 'desc'
@@ -46,7 +50,7 @@ module Resque
           reversed = true
         end
         all_items.each_with_index do |item, i|
-          if !class_name || (item['payload'] && item['payload']['class'] == class_name)
+          if !class_name || (item['payload'] && item['payload']['class'] == class_name && (original_limit -= 1) >= 0)
             if reversed
               id = (all_items.length - 1) + (offset - i)
             else
--- a/lib/resque/failure/base.rb
+++ b/lib/resque/failure/base.rb
@@ -53,15 +53,15 @@ module Resque
       # A URL where someone can go to view failures.
       def self.url
       end
-      
+
       # Clear all failure objects
       def self.clear(*args)
       end
-      
-      def self.requeue(index)
+
+      def self.requeue(*args)
       end
 
-      def self.remove(index)
+      def self.remove(*args)
       end
 
       # Logging!
--- a/lib/resque/failure/multiple.rb
+++ b/lib/resque/failure/multiple.rb
@@ -56,8 +56,8 @@ module Resque
         classes.first.requeue(*args)
       end
 
-      def self.remove(index, queue)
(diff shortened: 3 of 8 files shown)


lib/resque/failure/redis.rb
```diff
@@ -46,7 +50,7 @@ def self.each(offset = 0, limit = self.count, queue = :failed, class_name = nil,
           reversed = true
         end
         all_items.each_with_index do |item, i|
-          if !class_name || (item['payload'] && item['payload']['class'] == class_name)
+          if !class_name || (item['payload'] && item['payload']['class'] == class_name && (original_limit -= 1) >= 0)
```
chrisccerami: Just a minor comment, but is there a reason that I'm not seeing for using `orignal_limit -= 1` instead of just `original_limit - 1`?

lukeasrodgers:
This seems to work for me.

1-x-stable broken:
screen shot 2016-02-04 at 9 33 08 pm


With this fix:
screen shot 2016-02-04 at 9 34 13 pm

One thing to note is that this fix (perhaps unintentionally) changes to display of the range of failures, which results in some counterintuitive UI ("0 to 0 of 1 jobs"):

screen shot 2016-02-06 at 10 54 31 am

It's unclear to me how to address that. For example, saying "Showing 0 to 17 of 17 jobs" (which is the current UI on 1-x-stable) isn't totally accurate because those numbers represent indices into the failure array, and `failures[17]` doesn't exist. 

The best solution might be:
* 0 jobs: "No jobs"
* 1 job: "Showing 1 job"
* N jobs: "Showing 0 to N-1 jobs"
* N jobs where N is greater than page size (of 20, for example): "Showing 20-39 of 55 jobs"

I realize that some of these issues are somewhat unrelated to this particular fix, but I think it would be good to address them together.

lib/resque/failure/redis.rb
hasaniskandar: `original_limit` acts as a counter here, `original_limit - 1` will produce the same result on each loop.

hasaniskandar:
Thanks @lukeasrodgers for pointing that out. I will take a look when I have a chance.

chrisccerami:
@hasaniskandar Things look good for when there is 1 failed job:

screen shot 2016-04-01 at 9 21 57 pm


but still look incorrect when there is more than one (in this case there are 2 failed jobs):

screen shot 2016-04-01 at 9 23 15 pm

hasaniskandar:
@chrisccerami Thanks for the feedback. First, it is correct per @lukeasrodgers's suggestion:

> 1 job: "Showing 1 job"
> N jobs: "Showing 0 to N-1 jobs"

Second, I agree with you that the pagination is weird indeed, but thats a different issue. Could you please create a separate issue or PR instead? It would be great if you can help that way.

The original purpose of this PR is to fix failed jobs that do not display a complete list when `params[:class]` provided. I would like this PR to be merged ASAP so I can point back my project to the original resque repository.

chrisccerami:
@hasaniskandar sure, I think that's reasonable. Let me open a related issue to remind myself and others of this and then I'll most likely merge this after another quick review of the code. Sorry that so much of this UI cleanup got put on you when you were trying to solve a simple problem, it's easy to let scope creep like that.

chrisccerami:
@hasaniskandar I don't think I've got any issues with this code overall. My only remaining request would be if you could add messages to your assertions in the tests just to make things clearer? I find it tends to help whoever needs to work on those tests next if they know what each assertion is really testing and/or why.

hasaniskandar:
@chrisccerami Great, thanks! Yes sure, I will add messages to the assertions in the tests to make it clearer.

Issue #1299, opened 2015-05-08
Log errors with the appropriate (ERROR) log level


bob-rove:
Currently, all the messages are logged with DEBUG or INFO log level.
So in case of error, exception will be logged with INFO level. And if you are following a good practice to log only WARN and ERROR messages in production, you are in trouble. There is no way to resolve any issues in production with Resque failures before begin..resque block (https://github.com/resque/resque/blob/v1.25.2/lib/resque/worker.rb#L186) as you won't see anything in your logs.
As an example, if you have a short-term problem with connectivity between Resque and Redis, then worker will die trying to reserve the job and you will se nothing in the logs (if your verbosity level WARN).

chrisccerami:
I can take a crack at this.

chrisccerami:
@steveklabnik The `CONTRIBUTING.md` mentions `CHANGELOG` entries, but I don't see a Changelog in the repo. Am I missing something, or is the Contributing file out of date?

steveklabnik:
It's that we lost the old CHANGELOG because I'm a bad maintainer.

I would like to have a good CHANGELOG, it was one of the things I was going to investigate for the next release.

chrisccerami:
@bob-rove this should be addressed now by #1405

PR #1405, opened 2016-02-06
Log errors at the correct log level

chrisccerami:
This should address #1299. I'm more familiar with RSpec for testing so I apologize in advance if the tests aren't stellar. I also saw in `CONTRIBUTING.md` that changes should be listed in the Changelog, which didn't exist, so I figured what better time than now? Let me know if you don't want that in this PR, or if you'd like it in a different format.


lib/resque/worker.rb
```diff
@@ -813,17 +813,11 @@ def solaris_worker_pids
...
 
-    # Log a message to Resque.logger
-    # can't use alias_method since info/debug are private methods
-    def log(message)
-      info(message)
-    end
-
-    def log!(message)
-      debug(message)
+    def log(severity, message)
```
steveklabnik: So, I like this change, but this bit of information is swapped out at the moment: was this a public interface? By changing this, do we break things? I'm not sure off the top of my head. I don't have time to dig into this right now, but that's the only question I have about this PR, otherwise looks great :+1: 

(I am too used to Rust and static typing and am now slightly paranoid :sweat:)

chrisccerami: That paranoia is a good thing, and I think you're right. I'm pretty sure this actually is a public signature. I'll adjust it in a bit. Good catch!

lib/resque/worker.rb
```diff
@@ -858,5 +857,11 @@ def very_verbose=(value)
 
       @very_verbose = value
     end
+
+    private
+
+    def log_with_severity(severity, message)
```
chrisccerami: This should maintain the public interface of `log` and `log!` in case people are using them, but they're no longer used internally by `Resque::Worker`. There is still one caller to `log` in the `work` task of `tasks.rb`.

chrisccerami:
I should have stopped to think for a second when I was changing callers in the tests to `Worker#log`. Thanks for saving me that embarassment @steveklabnik! :rainbow:

lib/resque/worker.rb
steveklabnik: I wondered which approach you were gonna take. I didn't even think of this one. :+1:

steveklabnik:
Thanks for the patch!

PR #1408, opened 2016-02-07
Deprecate repeated methods in Resque stdlib.

maclover7:
All of these methods are all provided by ActiveSupport, and will be removed in 2.0. In v2.0, the methods in Resque::Helpers should be converted to use `ActiveSupport`. These methods should be standardized in one location, and not spread out throughout the whole codebase.

cc @steveklabnik

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -5,6 +5,9 @@
  * Add support for RESQUE_PROCLINE_PREFIX environment variable to prefix procline
    strings with a static identifier. (@rbroemeling)
 * Resque::Worker logs errors at the correct logging level (@chrisccerami)
+  * Deprecate repeated methods in Resque stdlib. These methods are all
+    provided by ActiveSupport, and will be removed in 2.0
+(@maclover7)
 
 ## 1.25.2 (2014-3-3)
 
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -1,6 +1,7 @@
 require 'mono_logger'
 require 'redis/namespace'
 require 'forwardable'
+require 'active_support/deprecation'
 
 require 'resque/version'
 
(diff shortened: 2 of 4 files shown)


steveklabnik:
So, I like this patch in theory, but at the same time, it also adds the ActiveSupport dependency. I'm not sure how I feel about that, exactly. 

Plus, saying "in 2.0" is also a bit weird, given that I'm not sure if and when a 2.0 will ever happen.

So, let me think about it :)

steveklabnik:
I am going to punt on this until the next release, but I think I will probably merge it in.

chrisccerami:
@steveklabnik same comment here about the release having been cut. I'm also not sure how I feel about adding the ActiveSupport dependency. On one hand, I would bet that the majority of Resque users are using it with Rails, so it's not the worst dependency to add. On the other, any dependency obviously comes at some cost. I think this is innocuous enough though.

chrisccerami:
As I've been thinking about this, it's starting to make less sense to me, although I'm still interested in @steveklabnik's opinion. Basically, without a 2.0 in sight, deprecation seems weird because that means we're not planning any breaking releases in the near future. This would also mean adding an ActiveSupport dependency which, while probably a not big deal for most, is also an added dependency which doesn't really get us much benefit. I guess unless there's a compelling reason for this change, I'm :-1: on this. Just my thoughts though.

steveklabnik:
> Basically, without a 2.0 in sight, deprecation seems weird because that means we're not planning any breaking releases in the near future.

I do think it's worthwhile to use deprecations to suggest what you _should_ be doing, even if there's no plans to eventually remove. In this specific case, getting people to stop using `Resque::Helpers` is a good idea, because it makes it really tough to have more than one instance of Resque working.

But, as always, it depends. Without a 2.0, this also means that there's less of a reason to do it overall, as you've said. So, it really just depends.

fw42:
Can we close this PR? Or is this still being worked on?

PR #1409, opened 2016-02-08
Add SIGHUP because golang (and therefore docker) can't forward SIGCONT

fw42:
This is a (rebased) cherry-pick of https://github.com/Shopify/resque/commit/96f4ddb80f5842d6f43323f0076a2f9cdcb22b33 and a work-around for https://github.com/golang/go/issues/8953 (https://go-review.googlesource.com/#/c/18185/).

This PR effectively makes SIGHUP an alias of SIGCONT, so that we can actually use this functionality with Docker. This can be removed at some point in the future (who knows when), but since the upstream golang change is pretty recent, I think it might be worth it to merge it for now.

Ping @burke @steveklabnik @dylanahsmith for review.

Probably should ping someone (?) from Docker in case anyone cares or has similar problems with other applications that also want to use SIGCONT? cc @jfrazelle

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -435,6 +435,7 @@ Resque workers respond to a few different signals:
 * `USR1` - Immediately kill child but don't exit
 * `USR2` - Don't start to process any new jobs
 * `CONT` - Start to process new jobs again after a USR2
+* `HUP` - Start to process new jobs again after a USR2
 
 If you want to gracefully shutdown a Resque worker, use `QUIT`.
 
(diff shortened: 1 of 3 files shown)


steveklabnik:
Tweeted for some exposure, dunno how much attention that will get.

ecoffey:
The patch looks clean and straightforward.

The only nervous part is you effectively bake into the public API of a resque worker a go/docker workaround.

I think the important question to answer before hitting "merge" is do we ever want to use SIGHUP for something else?

calavera:
`Docker maintainer here`

This looks like a good temporary patch. Go 1.6 will include the mentioned fix in the language, but we won't use it in a release until our next major one, v1.11, probably about April.

I agree with @ecoffey though.

dylanahsmith:
I think we should just recommend a workaround.  I think the simplest workaround is to setup a signal handler to forward the signal, which could be done in the resque:setup rake task:

```ruby
task "resque:setup" => :environment do
  trap('HUP') { Process.kill('CONT', Process.pid) } # remove once go and docker supports SIGCONT
end
```

fw42:
@dylanahsmith: Good idea. I like that much better. Closing here.

steveklabnik:
:+1: 

Maybe this should go in the README for the next few months?

PR #1410, opened 2016-02-08
Kill and abort buttons

fw42:
This cherry-picks https://github.com/Shopify/resque/pull/1 and https://github.com/Shopify/resque/pull/8 from Shopify's fork of resque.

The point of this PR is to add those two buttons:
screen shot 2016-02-08 at 18 22 03

Thoughts?

@steveklabnik @dylanahsmith 
cc @csfrancis @jpittis

### DIFF
--- a/lib/resque/server/views/workers.erb
+++ b/lib/resque/server/views/workers.erb
@@ -29,7 +29,16 @@
       <td class='process'>
         <% data = worker.processing || {} %>
         <% if data['queue'] %>
+          <% kill_worker = worker.terminal? %>
+          <% unless kill_worker %>
+            <form id="kill-worker-form" method="POST" style="display:none;"><input type="hidden" name="_method" value="DELETE"></form>
+          <% end %>
           <code><%= data['payload']['class'] %></code>
+          <% unless kill_worker %>
+            <a class="kill-tag" href="#" onclick="if(window.confirm('Are you SURE you want to kill this job?')) $('#kill-worker-form').submit()">Kill</a>
+          <% else %>
+            <b class="kill-tag">Killing</b>
+          <% end %>
           <small><a class="queue time" href="<%=u "/working/#{worker}" %>"><%= data['run_at'] %></a></small>
         <% else %>
           <span class='waiting'>Waiting for a job...</span>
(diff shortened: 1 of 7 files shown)


steveklabnik:
I am generally :+1: to this feature overall. Seems like the failure was spurious, maybe, so I'm gonna restart it.

fw42:
Hm test fail might be legit. Let me look into that.

chrisccerami:
I am definitely :+1: on this feature, as it's something we were discussing implementing for our own use of Resque. Haven't had a chance to really review to code yet, but I think it's a positive addition to Resque in general.

steveklabnik:
Merging the leaky text fix made this now have conflicts.

fw42:
Yeah I'm still looking at this. Please don't try to merge yet.

lib/resque/worker.rb
```diff
@@ -180,6 +180,52 @@ def glob_match(pattern)
...
+    def terminal?
+      redis.exists(kill_key)
+    end
+
+    def retry?
+      redis.get(kill_key) == "retry"
+    end
+
+    def kill!
+      redis.setex(kill_key, 300, "kill")
```
dylanahsmith: The UI talks about killing killing a job or its attempt, but we don't store an identifier for the job so that we don't accidentally kill the wrong job.  We intended to do that in Shopify, but this initial version never did that.

lib/resque/server/views/workers.erb
```diff
@@ -29,7 +29,18 @@
...
         <% data = worker.processing || {} %>
         <% if data['queue'] %>
+          <% kill_worker = worker.terminal? %>
+          <% unless kill_worker %>
+            <form id="retry-worker-form" method="POST" style="display:none;"><input type="hidden" name="_method" value="POST"></form>
+            <form id="kill-worker-form" method="POST" style="display:none;"><input type="hidden" name="_method" value="DELETE"></form>
+          <% end %>
           <code><%= data['payload']['class'] %></code>
+          <% unless kill_worker %>
+            <a class="kill-tag" href="#" onclick="if(window.confirm('Are you SURE you want to abort this job\'s attempt?')) $('#retry-worker-form').submit()">Abort Attempt</a>
```
dylanahsmith: Isn't retrying something that is added by [resque-retry](https://rubygems.org/gems/resque-retry).  I don't think resque itself has that concept.

fw42:
I'm closing this for now.. I'm really not happy with the code quality and the whole approach.

csfrancis:
The flaky tests were caused by the same issue that was fixed in https://github.com/ruby/ruby/pull/874.

PR #1380, opened 2016-01-13
Worker heartbeat-based cleanup (2nd try)

fw42:
This is another try at https://github.com/resque/resque/pull/1207 (basically cherry-picked https://github.com/Shopify/resque/commit/cc245ca61a9621e2a6f46defdc2095d3f8ca5f3b with a bit of cleaning up).

@sirupsen @dylanahsmith @steveklabnik

PR #1413, opened 2016-02-11
Fix leaky tests and use proper stubs

fw42:
https://github.com/resque/resque/pull/1380 introduced a background thread to the Worker class in order to do heartbeats. However, this thread is sometimes not cleaned up in the tests (only if `unregister_worker` is called) and keeps running, thus potentially leaking state into other tests or causing deadlocks (we saw this already while trying to figure out why the tests in https://github.com/resque/resque/pull/1410 fail sporadically).

This PR fixes that by always killing the threads after each test. Not the most bullet-proof method but works for now (suggestions welcome). Also, replace some ghetto instance method monkey-patching with a simple `stubs`.

@csfrancis @steveklabnik, please review

### DIFF
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -24,6 +24,10 @@ describe "Resque::Worker" do
     Resque::Job.create(:jobs, SomeJob, 20, '/tmp')
   end
 
+  after do
+    @worker.kill_background_threads
+  end
+
   it "can fail jobs" do
     Resque::Job.create(:jobs, BadJob)
     @worker.work(0)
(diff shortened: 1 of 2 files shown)


test/worker_test.rb
```diff
@@ -933,62 +937,31 @@ def run_in_job(&block)
...
-
-          def reconnect
-            raise Redis::BaseConnectionError
-          end
-        end
-
-        def @worker.sleep(duration = nil)
-          # noop
-        end
+      @worker.redis.client.stubs(:reconnect).raises(Redis::BaseConnectionError)
```
steveklabnik: :clap: :clap: :clap: :clap:

steveklabnik:
> Not the most bullet-proof method but works for now (suggestions welcome). 

Nothing _immediately_ comes to mind, and it's certainly an improvement, even though I agree it feels like there might be something cleaner hiding in there. But I'm happy with incremental improvement.

PR #1414, opened 2016-02-11
Increase verbosity for DirtyExit

fw42:
Just some small changes that helped us with debugging a production issue.

@pushrax @steveklabnik


pushrax:
LGTM

steveklabnik:
I like verbosity :+1: 


Well, at least in this way.

PR #1418, opened 2016-02-18
Improve the command to determine the worker pids.

jroith:
The old command also returns the pids of other processes. In particular, it returns the pid of resque-scheduler (https://github.com/resque/resque-scheduler) and of rake tasks with a "resque:" prefix.

This change additionally checks for a dash and a number (the version number) to fix the problem.


steveklabnik:
Thank you! Looks great.

gabetax:
The `grep -v "resque-web"` portion is redundant with this change, correct?

PR #1419, opened 2016-02-20
Exclude files in `test` directory for Coveralls test coverage

RochesterinNYC:
Coveralls seems to be analyzing test coverage of files in the `test` directory and this should not be the case. This PR excludes these files from coverage analysis.


steveklabnik:
Nice catch! Thank you.

PR #1420, opened 2016-02-22
Avoid N+1 query by caching job in Worker instance

fw42:
This is essentially https://github.com/Shopify/resque/pull/11.

The problem that we run into is that we call `Reqsue::Worker.working` (at which point we already know all of the jobs, since they are part of the serialization that we already fetched from the Redis server). Then, later, for each of those working workers, we call `#job`, which fetches the current job again with an additional Redis call.

This PR adds an optional flag to the `#job` method that allows you to avoid this kind of N+1 query and use the cached previous value. The default is to not do that, so this should be backwards compatible.

Thoughts?

@Sirupsen @steveklabnik

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -78,7 +78,9 @@ module Resque
       end
 
       reportedly_working.keys.map do |key|
-        find(key.sub("worker:", ''), :skip_exists => true)
+        worker = find(key.sub("worker:", ''), :skip_exists => true)
+        worker.job = worker.decode(reportedly_working[key])
+        worker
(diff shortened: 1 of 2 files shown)


steveklabnik:
Seems pretty reasonable at first glance. I'm not sure if there's a better way to do this, but adding an optional param seems like a very small change, even if it's not the best possible way.

Sirupsen:
I'd love for this to make it in. Our real-life use case is that the amount of network round-trips just doesn't scale.

steveklabnik:
let's do it then!

PR #1425, opened 2016-03-09
add an option: `Resque.enqueue_front = true`

nathantsoi:
adds an option to push new jobs to the beginning of the queue

useful in instances where we want LIFO job execution instead of the default FIFO

the default value `Resque.enqueue_front = false` will push new jobs to the end of the queue

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -165,6 +165,12 @@ module Resque
     @prune_interval || DEFAULT_PRUNE_INTERVAL
   end
 
+  attr_writer :enqueue_front
+  def enqueue_front
+    return @enqueue_front unless @enqueue_front.nil?
+    @enqueue_front = false
+  end
+
   # The `before_first_fork` hook will be run in the **parent** process
   # only once, before forking to run the first job. Be careful- any
   # changes you make will be permanent for the lifespan of the
@@ -267,7 +273,7 @@ module Resque
     encoded = encode(item)
     redis.pipelined do
       watch_queue(queue)
-      redis.rpush "queue:#{queue}", encoded
+      redis.send((enqueue_front ? 'lpush' : 'rpush'), "queue:#{queue}", encoded)
     end
   end
 


nathantsoi:
force push for ruby 1.8.x compatibility

lib/resque.rb
```diff
@@ -165,6 +165,12 @@ def prune_interval
     @prune_interval || DEFAULT_PRUNE_INTERVAL
   end
 
+  attr_writer :enqueue_front
+  def enqueue_front
+    return @enqueue_front unless @enqueue_front.nil?
+    @enqueue_front = false
+  end
```
dimko: Why not just use `attr_accessor` for that?

nathantsoi: I was just following the format of the existing code (https://github.com/resque/resque/pull/1425/files#diff-bb00e7d239ca836ced064c1f97814e23L159)

I guess this also allows setting a default value? If there is a better way, I'm down to implement that?

dimko: I mean `attr_accessor :enqueue_front` will do the same thing.

nathantsoi:
not sure I follow, would that replace the `attr_writer` line or the method as well? if it's the latter, how would we set the default value?

lib/resque.rb
nathantsoi: not sure I follow, would this replace `attr_writer` or both the `attr_writer` and enqueue_front method? if it's the latter, where do we set the default `@enqueue_front` value?

steveklabnik: Yes, `attr_accessor` is not the same here, as it doesn't have a default.

steveklabnik:
Thanks for the patch @nathantsoi !

I'm going to be cutting a Resque release very soon, which has already undergone some testing, and don't like to merge a new feature right before cutting a release. So I'm going to wait to merge until after the release, and we can get this in the one after.

lib/resque.rb
dimko: Yea, it will be `nil`. But in this case (`enqueue_front ? ... : ...`) it will do the same thing. Anyway, it's up to you.

steveklabnik: Ah yes, I see. It's the same thing in the sense that where this is _used_, it has the same effect, not that it's the same at this line of code. :+1:

steveklabnik: (In that case, I am not sure which approach I prefer. I might even lean towards the attr_accessor

nathantsoi: ahh, i see. yep either way works for me!

nathantsoi:
rad, thx @steveklabnik

chrisccerami:
@steveklabnik I believe the new release has been cut, yes?

steveklabnik:
It has, yes!

nathantsoi:
Awesome. Thx all!

dylanahsmith:
@chrisccerami your PR https://github.com/resque/resque/pull/1466 that adds Resque::DataStore made Resque.enqueue_front get ignored.  Should we remove `Resque.enqueue_front` and instead suggest that a custom Resque::DataStore be used instead?  Or should we continue supporting return support for Resque.enqueue_front?

It doesn't look like this has actually been released, which is confusing given the talk of a release being cut after this

chrisccerami:
This PR was merged April 2 and the most recent release was cute in March.

dylanahsmith:
on the latest release, 1.26.0, lib/resque.rb doesn't have enqueue_front https://github.com/resque/resque/blob/v1.26.0/lib/resque.rb

chrisccerami:
Right, 1.26.0 was tagged in March. This PR that added `enqueue_front` was merged in April.

dylanahsmith:
Right, so do we actually want this?  If so it is no longer has any effect

TSMMark:
It is a shame this was cut. I would like this feature.

chrisccerami:
I don't know what you mean @TSMMark. This pull request was merged and has since been released in 1.27.0 on February 8, 2017.

PR #1427, opened 2016-03-10
Fix regression

steveklabnik:
reported here: https://github.com/resque/resque/pull/1374#issuecomment-172924862

This seems like a straightforward fix: the methods had different signatures. Since this proxies everything to the first backend, adding the param and ignoring it _should_ fix it, I think.

@lukeasrodgers can you maybe try with this patch and see if it fixes said regression? That would be very awesome. I' going to do my own investigation as well...

### DIFF
--- a/lib/resque/failure/multiple.rb
+++ b/lib/resque/failure/multiple.rb
@@ -56,7 +56,7 @@ module Resque
         classes.first.requeue(*args)
       end
 
-      def self.remove(index)
+      def self.remove(index, queue)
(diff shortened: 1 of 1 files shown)


lukeasrodgers:
@steveklabnik yes this looks good to me, and I tested it on an app that uses resque's multiple failure backend and it appears to work as intended (both with the built-in resque web UI as well as a 3rd-party failure management plugin).

steveklabnik:
Great! Let's get it merged then. :confetti_ball:

Issue #1428, opened 2016-03-11
Airbrake handler calls .notify_or_ignore, which was removed from the airbrake gem


seanwalbran:
[This code](https://github.com/resque/resque/blob/0debc8193d7a26715cd55aa382b18ab5695fab6f/lib/resque/failure/airbrake.rb#L24) calls Airbrake.notify_or_ignore, which was [removed in the 5.x versions of the Airbrake gem](https://github.com/airbrake/airbrake/blob/master/docs/Migration_guide_from_v4_to_v5.md#notify-or-ignore), in favor of simply `Airbrake.notify`.

steveklabnik:
We've been wanting to remove the airbrake integration and make it a separate gem anyway.

kitop:
@steveklabnik I'm glad to help here, as I have the same issue.
Would you prefer a PR fixing this (to account for both cases) or to create a separate gem now?

steveklabnik:
Whichever works for you, frankly. Thanks so much for offering to pitch in!

kitop:
Oh, actually the new airbrake already seems to provide this: https://github.com/airbrake/airbrake/blob/master/lib/airbrake/resque/failure.rb

The tricky part, I guess, is how to tell your app to require that one and not the one provided by resque when you have airbrake >= 5.

kitop:
Oh, spoke too soon. It's actually documented: https://github.com/airbrake/airbrake/blob/master/docs/Migration_guide_from_v4_to_v5.md#resque

I think that solves the issue, without doing anything :)
@seanwalbran does that work for you?

seanwalbran:
Looks like it, thanks!

steveklabnik:
Oh rad, even better :)

> The tricky part, I guess, is how to tell your app to require that one and not the one provided by resque when you have airbrake >= 5. 

I was planning on saying "This release removes airbrake support, but please drop in this gem and it will work". So if it's functionally equivalent, we can just do that.

PR #1429, opened 2016-03-11
notify_or_ignore is gone in v5

orenmazor:
Updated my airbrake earlier this week and my jobs all immediately started failing due to this change: https://github.com/airbrake/airbrake/blob/23457a14144013959ffc4f54683238929e30f908/docs/Migration_guide_from_v4_to_v5.md#notify-or-ignore

### DIFF
--- a/lib/resque/failure/airbrake.rb
+++ b/lib/resque/failure/airbrake.rb
@@ -21,7 +21,7 @@ module Resque
       end
 
       def save
-        ::Airbrake.notify_or_ignore(exception,
+        ::Airbrake.notify(exception,
             :parameters => {
             :payload_class => payload['class'].to_s,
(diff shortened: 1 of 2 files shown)


crazymykl:
Please fix the test failure on 1.8.7.

orenmazor:
@crazymykl donezo :+1:

chrisccerami:
I think there is a plan to remove Airbrake from Resque itself and instead use the separate gem. I'm not entirely sure what the status of that work is, but I assume this change would be compatible? @orenmazor

steveklabnik:
It's true, but that work hasn't been done, so it's useful to merge in the meantime. Thanks @orenmazor !

chrisccerami:
I believe this fixes #1428, right @steveklabnik?

PR #1430, opened 2016-03-14
adds rails 4 installation instruction

edshadi:
This is a small addition, but I believe that if you're new to the gem (I'm not) and you don't see a Rails 4 installation, you'll probably move on.


steveklabnik:
Thank you! I agree wholeheartedly.

PR #1434, opened 2016-03-19
Workers queried out now have the correct hostname

### LINKED ISSUE #1304: Misleading worker hostname implementation
Current implementation assumes that instance of Worker class is always used on same machine:
```ruby
# lib/resque/worker.rb
def hostname
  Socket.gethostname
end
```
When we have worker instances on worker machines (like ALL heroku deployments), then running rails console:
`Resque.workers[0].hostname` will return hostname of rails console machine

It seems wrong to invoke runtime-dependent class method, and to expose it as instance method

crazymykl:
Fixes #1304

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -48,6 +48,7 @@ module Resque
     # registered in the application. Otherwise, forked workers exit with `exit!`
     attr_accessor :run_at_exit_hooks
 
+    attr_writer :hostname
     attr_writer :to_s
     attr_writer :pid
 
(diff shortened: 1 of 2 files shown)


chrisccerami:
:+1: This looks like it solves #1304 to me. Definitely preferential to the previous behavior.

chrisccerami:
I pulled this branch down and ran it locally with some breakpoints, just to clarify for myself that MiniTest stubbing works the way I expected, and everything seems good to me :+1:

PR #1435, opened 2016-03-20
Add support for removing a failure queue to Redis::Failure::Multiple

tneems:


chrisccerami:
Hey @tneems, I think some tests would be greatly appreciated as this is a new public method.

PR #1437, opened 2016-03-30
Восстановление workers

ff40:


chrisccerami:
@ff40 I think this PR has an issue. What I suspect has happened is that the `vladimirich:master` branch is based on the old `2.0` version of master, whereas current `resque:master` is the recently renamed `1-x-stable`, although @steveklabnik may want to correct me on that if I'm wrong. It's not clear to me based on the title (or the description, since there is none) what this PR is meant to do, but if there is a change you do wish to have merged into current master, it needs to be based off of what was previously `1-x-stable`.

steveklabnik:
Yes, that seems to be correct. Sorry about that @ff40 !

PR #1438, opened 2016-03-31
Fix race condition on worker startup

stevedomin:
Commit 6a0746ccd7307c2caebfe8cb6170d8f8c0c55067 improved detection of
worker pids on the same host but was missing one of the valid state
a worker can be in on startup. As a result sometimes the first worker
would be unregistered by other workers starting shortly after.

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -797,7 +797,7 @@ module Resque
     # Find Resque worker pids on Linux and OS X.
     #
     def linux_worker_pids
-      `ps -A -o pid,command | grep -E "[r]esque:work|[r]esque-[0-9]" | grep -v "resque-web"`.split("\n").map do |line|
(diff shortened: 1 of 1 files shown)


stevedomin:
I'm not sure what caused the build to fail, seems to be unrelated from my change.

chrisccerami:
@stevedomin I'm also not seeing that failure locally when running the tests on `1.9.3`. I'm wondering if it would pass if the build was rerun.

steveklabnik:
I restarted the build and it does, in fact, work!

gabetax:
@steveklabnik I didn't see this in CONTRIBUTING.md: what's the workflow on cherry-picking this into the 1-x-stable branch? I have resque plugin that's misbehaving on 1.26.0 due to this regression since #1418 was merged directly into 1-x-stable and this is into master.

PR #1439, opened 2016-03-31
worker: do not trigger verbose logging if env variables are not set

ldnunes:
When setting a custom logger the logger state is lost due to a regression added in the following commit:

,----
| commit 87ae6d5e1cd6ebc498dc0a04ada74dc770c7facf
| Author: Jacob Green <jacob@nationbuilder.com>
| Date:   Mon Jan 25 11:22:17 2016 -0500
|
|     Refactor worker#initialize
`----

The verbose and very_verbose setter methods are always called, even when no verbose environment variables are set. This makes a custom logger loose its level or custom formatter.

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -127,8 +127,9 @@ module Resque
       @paused = nil
       @before_first_fork_hook_ran = false
 
-      self.verbose = ENV['LOGGING'] || ENV['VERBOSE']
-      self.very_verbose = ENV['VVERBOSE']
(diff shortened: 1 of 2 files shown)


chrisccerami:
:+1: I appreciate seeing new unit tests being added when a regression is caught.

steveklabnik:
Gah! Thank you!

Also, I will manually merge this into master, but please send PRs to master these days, I've switched what was 1-x-stable over to it.

PR #1447, opened 2016-04-15
Prompt for confirmation on clearing failed jobs

neslom:
The `Retry Failed Jobs` and `Clear Failed Jobs` buttons are very close to each other and do very different things, so I think it would be nice to prompt for confirmation when clearing jobs to make sure that's indeed what the user is intending to do. 😄

### DIFF
--- a/lib/resque/server/views/failed.erb
+++ b/lib/resque/server/views/failed.erb
@@ -6,7 +6,7 @@
 
 <% unless failed_size.zero? %>
 <form method="POST" action="<%= u "failed#{'/' + params[:queue] if params[:queue]}/clear" %>">
-  <input type="submit" name="" value="Clear <%= params[:queue] ? "'#{params[:queue]}'" : 'Failed' %> Jobs" />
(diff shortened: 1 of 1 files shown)


chrisccerami:
This seems like a sensible change. Do you think it also makes sense to add the confirmation to the "Retry Failed Jobs" button?

chrisccerami:
@neslom Also, could please reopen this PR against `master` rather than `1-x-stable`? Sorry for the inconvenience.

neslom:
No problem @chrisccerami. Should all PRs be against `master` at this point? I can update the CONTRIBUTING as well if this is the case.

I only included the prompt for clearing jobs because it seemed to be more harmful than retrying jobs accidentally, but I'd be happy to include it for both if you think it's a good idea.

Let me know what you think and then I'll open this again!

chrisccerami:
@neslom Yeah, `1-x-stable` has been renamed back to `master` now that plans for 2.0 have been scrapped. The reason I'm thinking confirmation on retrying jobs might be valuable is that some background jobs are very memory intensive (hence being backgrounded) and someone retrying a large number of failed jobs all at once accidentally could be problematic, but at the same time, they should have the correct number of workers handling that queue so as not to overwork themselves.

I think your PR is fine as is if you just switch it to `master`.

Issue #1446, opened 2016-04-14
Calling Worker.all or Worker.working causes workers to overwrite PIDFILE


jeremywadsack:
We just upgraded from resque 1.25.2 to 1.26.0 and suddenly some of our workers are overwriting the PIDFILE with each fork:

```
$ cat tmp/pids/resque_worker_prefetcher.pid
17278
$ ps aux | grep [1]7278
deploy   17278  4.5  0.7 605992 180332 ?       Sl   10:55   0:01 resque-1.26.0: Processing prefetcher since 1460656561 [MinimumPrefetcher]                                                                                                                        
deploy   31520  1.0  0.6 588980 171968 ?       Sl   10:15   0:25 resque-1.26.0: Forked 17278 at 1460656561                                                                                                                                                        
```

In this example, the recorded PID should be 31520, not 17278.

As we use monit to manage our workers this caused monit to continually launch new workers each time the fork terminated (after finishing the job). 

What's troubling me is that this only happens for some workers. I _think_ it's workers for which the queue has jobs in it when the workers start. Although it also seems to also be limited to jobs that run quickly (under a minute). This is making it hard for me to debug this, much less come up with a specific reproduction (but I will work on that).

I see that there was a [change to where PIDFILE was written](https://github.com/resque/resque/commit/7d0b20ee00986b749b03d4407af199039b504cc2) between 1.25.2 and 1.26.0, however, the code to write the pid is in the `initialize` method which shouldn't be called after `fork` which is called from `#work`. So I'm confused as to how this could even happen.

We are not using `BACKGROUND` and are depending on the shell (`&`) to background the workers.

I am going to keep digging into this to come up with a reproduction of the issue but I wanted to post this here in case anyone else has seen it or has thoughts of where I can look further for how this would happen.

jeremywadsack:
I've narrowed this down to the use of `Worker.all` or `Worker.working`.

This only happens for some workers. Those workers are using an "inspector" class we wrote to check on the state of existing workers. This called `Reque.all` and `Reque::Worker.working`.

The code in those methods [call `#new` on the worker](https://github.com/resque/resque/blob/0d5389cb21744f59993af8f24d855a20d0e7b9ae/lib/resque/worker.rb#L57) which rewrites the PID file. Because we are calling those class methods from _within_ the running fork, they write the PID file. In fact, they end up writing the PID file many many times as they create a new instance of the worker from the _current_ process for each registered worker.

[...]

PR #1448, opened 2016-04-15
Initializing a worker should not write the PID file or re-daemonize

jeremywadsack:
Class methods like `::find` (and those that depend on it, like `::all` and `::working`) call `#new` to create new instances of the worker objects. If these methods are called from within the forked object it would cause `#initialize` to rewrite the `PIDFILE` or (possibly?) re-daemonize the process. 

This resolves that by making a separate method called by the rake task when starting a worker.

See #1446. I have run this in our environment and looks like it resolves the issue, although I want to monitor it some more to ensure this is the correct fix.

I have two concerns about this and could use some suggestions from someone more familiar with the code base:

1. Is it a problem that we are now daemonizing (if `ENV["BACKGROUND"]` is set) _after_ we assign the queues? I don't use this approach so I'm not familiar with how it works.

2. I created the new `#prepare` method that handles these which requires the rake task to call that explicitly. I think this is a good idea because then the consumer can decide whether to perform these actions. I considered having the `#work` method automatically call `#prepare` if it had not been called before which would make this an implicit call. But I think that hides too much magic (which could lead to further issues like this), adds complexity, and delays these two actions later in the start up which could also be a problem. Feedback on my approach, though, is appreciated.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,9 @@
 
 None yet!
 
+* Fix issue where calling Worker.find, Worker.all, or Worker.working from withing
+  a running job would rewrite the PID file with the PID of the forked worker. (@jeremywadsack)
+
 ## 1.26.0 (2016-03-10)
 
 This changelog is a bit incomplete. We will be much stricter about the changelog for
--- a/lib/resque/tasks.rb
+++ b/lib/resque/tasks.rb
@@ -15,6 +15,7 @@ namespace :resque do
       abort "set QUEUE env var, e.g. $ QUEUE=critical,high rake resque:work"
     end
 
+    worker.prepare
     worker.log "Starting worker #{self}"
     worker.work(ENV['INTERVAL'] || 5) # interval, will block
   end
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -124,6 +124,9 @@ module Resque
     # If passed a single "*", this Worker will operate on all queues
     # in alphabetical order. Queues can be dynamically added or
     # removed without needing to restart workers using this method.
+    #
+    # Workers should have `#prepare` called after they are initialized
+    # if you are running work on the worker.
     def initialize(*queues)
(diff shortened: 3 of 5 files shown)


chrisccerami:
This does seem like it solves an actual problem that people are facing. The only request I have is that if this changes the way people need to handle workers, would you mind adding that info to the Readme or the wiki?

chrisccerami:
I'm going to merge this for the next release, as like I said, it does seem like it solves a problem people are actually facing right now, but if you can @jeremywadsack I would still really appreciate an update to the README and/or wiki if you feel its necessary.

jeremywadsack:
Chris, I'll try to get a Readme or wiki update today.

Thanks for merging!
On Thu, May 26, 2016 at 8:20 AM Chris C Cerami <notifications@github.com>
wrote:

> Merged #1448 <https://github.com/resque/resque/pull/1448>.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/resque/resque/pull/1448#event-673315702>
>
-- 
Jeremy Wadsack

jeremywadsack:
@chrisccerami I looked through the README and wiki and I don't see anything that would change based on the changes I made. What I changed is the implementation on how `Resque::Worker.new` works and I documented that in [the comments](https://github.com/resque/resque/pull/1448/commits/08340ca19c411a9b35b7e51c0232e3ea1cc72300#diff-693c73ae6c0943d273cdb8efc95ecec2R128) for that method. There's no change to the rake task or to the parameters to it, so I think we don't need to update anything there.

What I didn't check is whether any of the many plugins would break with this change. But only if they depend on the worker consuming `ENV["PIDFILE"]` or `ENV["BACKGROUND"]` and they bypass the rake task.

I've pushed another commit to this branch with additional comments in the `HISTORY.md` which I think should suffice. Feel free to reword or rewrite that, however, in whatever way makes the most sense.

PR #1454, opened 2016-04-21
Explicit require tilt/erubis.

rainerborene:
This change will make disappear the following warning message on threaded environments like Puma.
> WARN: tilt autoloading 'tilt/erubis' in a non thread-safe way; explicit require 'tilt/erubis' suggested.

### DIFF
--- a/lib/resque/server.rb
+++ b/lib/resque/server.rb
@@ -1,5 +1,5 @@
 require 'sinatra/base'
-require 'erb'
+require 'tilt/erb'
 require 'resque'
 require 'resque/version'
 require 'time'


rainerborene:
Travis is complaining about Ruby 1.8.7 compatibility. It's time to drop support for this version, right?

chrisccerami:
I don't think there are currently plans to stop supporting Ruby 1.8.7, but I also see this failing on the other Ruby versions as well.

rainerborene:
Hmm. Turns out `erubis` is an external dependency and requiring `tilt/erb` did the trick

PR #1457, opened 2016-04-26
Use flag to stop heartbeat thread rather than Thread.kill

Sinjo:
I was reading the code for heartbeats (which, btw, ❤️) and noticed that it uses `Thread.kill` to stop the heartbeat thread when Resque is exiting. While I've not observed a problem in production with this approach, I'm aware that it's [not a safe thing to do in general](http://blog.headius.com/2008/02/ruby-threadraise-threadkill-timeoutrb.html).

I figured it wouldn't hurt to switch to exiting the loop (and therefore the thread) when a variable is set, rather than calling `kill` on the thread.

I've tested this locally, and it seems to start/stop just fine.

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -504,9 +504,15 @@ module Resque
 
     def start_heartbeat
       heartbeat!
+
+      @stop_heartbeat_thread = false
       @heart = Thread.new do
-        loop do
-          sleep(Resque.heartbeat_interval)
+        until @stop_heartbeat_thread do
+          Resque.heartbeat_interval.times do
+            sleep(1)
+            break if @stop_heartbeat_thread
+          end
+
           heartbeat!
         end
       end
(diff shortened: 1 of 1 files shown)


chrisccerami:
This new code seems reasonable to me, but I'm not terribly familiar with Ruby's Thread. The post you link is from 2008, and I have no idea if the situation has improved in the last 8 years. I don't have any major objections, other than I would like to see us not try to reengineer our own version of `Thread.kill` unless necessary. Does anyone have any input on whether this change seems necessary?

chrisccerami:
I think I feel good about this change. I'm going to merge and it will probably be included in the next release. Thanks @Sinjo 🌈 🌟

Sinjo:
Awesome! Thanks for merging.

On your question: `Thread.kill` and `Thread.raise` just aren't safe to use on arbitrary code (the examples in Charles Nutter's article still hold). For what it's worth, Java went through this same thing and [ended up deprecating the methods](https://docs.oracle.com/javase/8/docs/technotes/guides/concurrency/threadPrimitiveDeprecation.html).

PR #1458, opened 2016-05-06
Add requeue_all to Redis::Failure::Multiple

ale7714:
We're using Redis::Failure::Multiple on our project and notice that after updating resque to 1.26.0, `resque/failed/requeue/all` fails. 
37631d28-1302-11e6-9979-19b48f248448

### DIFF
--- a/test/resque_failure_multiple_test.rb
+++ b/test/resque_failure_multiple_test.rb
@@ -1,0 +1,16 @@
+require 'test_helper'
+require 'resque/failure/multiple'
+
+describe "Resque::Failure::Multiple" do
+
+  it 'requeue_all and does not raise an exception' do
+    with_failure_backend(Resque::Failure::Multiple) do
+      Resque::Failure::Multiple.classes = [Resque::Failure::Redis]
+      exception = StandardError.exception('some error')
+      worker = Resque::Worker.new(:test)
+      payload = { "class" => Object, "args" => 3 }
+      Resque::Failure.create({:exception => exception, :worker => worker, :queue => "queue", :payload => payload})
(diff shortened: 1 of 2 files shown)


test/resque_failure_multiple_test.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  it 'it requeue_all and does not raise an exception' do
+    with_failure_backend(Resque::Failure::Multiple) do
+      Resque::Failure::Multiple.classes = [Resque::Failure::Redis]
+      exception = StandardError.exception('some error')
+      worker = Resque::Worker.new(:test)
+      queue = "queue"
+      payload = { "class" => Object, "args" => 3 }
+      Resque::Failure.create({:exception => exception, :worker => worker, :queue => queue, :payload => payload})
+      Resque::Failure.requeue_all # should not raise an error
```
chrisccerami: I was very confused by why the coverage decreased when you clearly added a test for this, but I think I figured it out. Shouldn't this line be `Resque::Failure::Multiple.requeue_all` since that's the method you added?

ale7714: You're right sorry that was a silly mistake 🙈, pushing right call. Thanks

ale7714:
@chrisccerami actually it keeps giving me a decrease 😅, maybe is because there are no test covering the other methods on `Resque::Failure::Multiple`? would that make sense?

ale7714:
Let me know if you think I should do something more to improve the code

chrisccerami:
@ale7714 no i think this seems reasonable. I wish i understood what's happening with coveralls, but I don't see it as a blocker.

PR #1461, opened 2016-05-20
Indicate license in gemspec

NobodysNightmare:
If I see it correctly, you are using the MIT license for the `resque` gem.

This PR indicates the license in the gemspec.


steveklabnik:
That's true, thanks so much!

PR #1465, opened 2016-05-26
Update HISTORY.md with details about API change for plugin authors.

jeremywadsack:
See #1448.

This adds more information about the change to the `Worker#new` API for plugin authors that may depend on `ENV["PIDFILE"]` or `ENV["BACKGROUD"]` behavior and are not using the rake task.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,7 +3,9 @@
 None yet!
 
 * Fix issue where calling Worker.find, Worker.all, or Worker.working from withing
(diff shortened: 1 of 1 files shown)


chrisccerami:
Hm, it's worrying that the build is failing when this PR obviously can't be the cause. I'll look into it.

chrisccerami:
Well, it passed on the second try. That's...disconcerting. Anyway, thanks for the update!

PR #1468, opened 2016-06-02
Update jQuery file to newest version

martinstuecklschwaiger:
@chrisccerami you updated the reference to jQuery in #1463, but not the file itself.


chrisccerami:
😅 So I did. Thanks!

PR #1469, opened 2016-06-02
Remove to_h function invocation

martinstuecklschwaiger:
The `to_h` function does not exist in Ruby 1.9 yet. 
This replaces it with some custom logic. 

I'm actually not a Ruby programmer, so please correct me if that's wrong.

### DIFF
--- a/lib/resque/server/views/working.erb
+++ b/lib/resque/server/views/working.erb
@@ -17,10 +17,11 @@
         <% queue = data['queue'] %>
         <td><a class="queue" href="<%=u "/queues/#{queue}" %>"><%= queue %></a></td>
         <td><span class="time"><%= data['run_at'] %></span></td>
+	<% payload = data.key?('payload') ? data['payload'] : {} %>
         <td>
-          <code><%= data['payload'].to_h['class'] %></code>
+          <code><%= payload.key?('class') ? payload['class'] : "—" %></code>
         </td>
-        <td><%=h data['payload'].to_h['args'].inspect %></td>
+        <td><%=h payload.key?('args') ? payload['args'].inspect : "—" %></td>
       </tr>
   </table>
 
--- a/lib/resque/server/views/working.erb
+++ b/lib/resque/server/views/working.erb
@@ -17,7 +17,7 @@
         <% queue = data['queue'] %>
         <td><a class="queue" href="<%=u "/queues/#{queue}" %>"><%= queue %></a></td>
         <td><span class="time"><%= data['run_at'] %></span></td>
-	<% payload = data.key?('payload') ? data['payload'] : {} %>
+        <% payload = data.key?('payload') ? data['payload'] : {} %>
         <td>
           <code><%= payload.key?('class') ? payload['class'] : "—" %></code>
         </td>


martinstuecklschwaiger:
Ping @fw42 @Sirupsen. 

This really is just 3 changed lines and fixes the detail view in Ruby 1.9.3 (which still has tier 1 support). The bug actually was introduced by https://github.com/resque/resque/pull/1358 not long ago. 

This bug can be reproduced by opening a job detail view in the browser: 

job-detail-view

lib/resque/server/views/working.erb
```diff
@@ -17,10 +17,11 @@
         <% queue = data['queue'] %>
         <td><a class="queue" href="<%=u "/queues/#{queue}" %>"><%= queue %></a></td>
         <td><span class="time"><%= data['run_at'] %></span></td>
+	<% payload = data.key?('payload') ? data['payload'] : {} %>
```
fw42: indentation here is wrong

lib/resque/server/views/working.erb
```diff
@@ -17,10 +17,11 @@
...
         <td>
-          <code><%= data['payload'].to_h['class'] %></code>
+          <code><%= payload.key?('class') ? payload['class'] : "—" %></code>
```
fw42: hm. I'm not sure I understand why this is equivalent... can you explain?

fw42: I assume `data['payload']` is an array? and that's why we are calling `to_h` on it?

martinstuecklschwaiger: Thank's for your help, @fw42. This really is my very first try to fix anything in Ruby. 

Actually `data` looks as following:
```
{
  "queue"=>"my_queue",
  "run_at"=>"Thu Jun 30 20:55:35 CEST 2016",
  "payload"=>{
    "class"=>"MyClass",
    "args"=>[["my", "args", "1", 2]],
    "id"=>"9387cec11cb81ae1210cc57739e363fc"
  }
}
``` 
Although `data` may also be `{}` (if there's no job running). The `to_h` was needed, to prevent an error with this empty hash. That's also why I first tried, if the payload actually exists.

fw42:
Ah ok. I get it now. 👍

lib/resque/server/views/working.erb
```diff
@@ -17,10 +17,11 @@
         <% queue = data['queue'] %>
         <td><a class="queue" href="<%=u "/queues/#{queue}" %>"><%= queue %></a></td>
         <td><span class="time"><%= data['run_at'] %></span></td>
+        <% payload = data.key?('payload') ? data['payload'] : {} %>
         <td>
-          <code><%= data['payload'].to_h['class'] %></code>
+          <code><%= payload.key?('class') ? payload['class'] : "—" %></code>
```
crazymykl: The older idiom for `foo.to_h` is `Hash[foo]`

fw42:
Thanks for your contribution!

martinstuecklschwaiger:
Thanks a lot for merging! 

Since this is a bugfix, would it be possible to make a minor release, like `1.26.1`?

PR #1470, opened 2016-06-02
Use fork with a block to avoid having a child process escape its code path

dylanahsmith:
@fw42 & @Sirupsen for review

## Problem

When `Kernel.fork` is used without a block, then there will be a race condition where an exception can occur from a signal that causes the child process to leave its code path and execute exceptions handlers for the parent process.

Additionally, the Resque::Worker#fork abstraction was combining code paths for the child process and the non-forking code path, which was making the code harder to reason about.

## Solution

Refactor Resque::Worker#work to make the code easier to read and understand by separating the three code paths (no forking, fork parent and fork child) and isolating the child process code path by using fork with a block.

I also moved the reconnect to redis into the Resque::Worker#perform method, since it is related to the after_fork hooks which are run from that method, and because it avoids the need for another rescue for job failures.

I also removed `will_fork?` since using it before the first fork will cause a confusing result.  Instead, there is a fork_per_jobs attribute that gets explicitly set to `false` in Resque::Worker#work so this fork fallback flow is more straightforward to understand and refactor.

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -53,6 +53,7 @@ module Resque
     # registered in the application. Otherwise, forked workers exit with `exit!`
     attr_accessor :run_at_exit_hooks
 
+    attr_writer :fork_per_job
     attr_writer :hostname
     attr_writer :to_s
     attr_writer :pid
@@ -226,33 +227,34 @@ module Resque
           working_on job
 
           procline "Processing #{job.queue} since #{Time.now.to_i} [#{job.payload_class_name}]"
-          if @child = fork(job)
-            srand # Reseeding
-            procline "Forked #{@child} at #{Time.now.to_i}"
+          if fork_per_job?
+            run_hook :before_fork, job
             begin
-              Process.waitpid(@child)
-            rescue SystemCallError
-              nil
+              @child = fork do
+                unregister_signal_handlers if term_child
+                perform(job, &block)
+                exit! unless run_at_exit_hooks
+              end
+            rescue NotImplementedError
+              @fork_per_job = false
             end
-            job.fail(DirtyExit.new("Child process received unhandled signal #{$?.stopsig}")) if $?.signaled?
-          else
-            unregister_signal_handlers if will_fork? && term_child
-            begin
-
-              reconnect if will_fork?
-              perform(job, &block)
-
-            rescue Exception => exception
-              report_failed_job(job,exception)
+            if @child
+              srand # Reseeding
+              procline "Forked #{@child} at #{Time.now.to_i}"
+              begin
+                Process.waitpid(@child)
+              rescue SystemCallError
+                nil
+              end
+              job.fail(DirtyExit.new("Child process received unhandled signal #{$?.stopsig}")) if $?.signaled?
+              @child = nil
             end
-
-            if will_fork?
-              run_at_exit_hooks ? exit : exit!
-            end
+          end
+          unless fork_per_job?
+            perform(job, &block)
           end
 
           done_working
-          @child = nil
         else
           break if interval.zero?
           log_with_severity :debug, "Sleeping for #{interval} seconds"
@@ -300,7 +302,10 @@ module Resque
     # Processes a given job in the child.
     def perform(job)
       begin
-        run_hook :after_fork, job if will_fork?
+        if fork_per_job?
+          reconnect
+          run_hook :after_fork, job
+        end
         job.perform
       rescue Object => e
         report_failed_job(job,e)
@@ -344,28 +349,6 @@ module Resque
           log_with_severity :error, "Error reconnecting to Redis; quitting"
           raise
         end
-      end
-    end
-
-    # Not every platform supports fork. Here we do our magic to
-    # determine if yours does.
-    def fork(job)
-      return unless will_fork?
-
-      # Only run before_fork hooks if we're actually going to fork
-      # (after checking will_fork?)
-      run_hook :before_fork, job
-
-      begin
-        # IronRuby doesn't support `Kernel.fork` yet
-        if Kernel.respond_to?(:fork)
-          Kernel.fork
-        else
-          raise NotImplementedError
-        end
-      rescue NotImplementedError
-        @cant_fork = true
-        nil
       end
     end
 
@@ -746,12 +729,9 @@ module Resque
       state == :idle
     end
 
-    def will_fork?
-      !@cant_fork && fork_per_job?
-    end
-
     def fork_per_job?
-      ENV["FORK_PER_JOB"] != 'false'
+      return @fork_per_job if defined?(@fork_per_job)
+      @fork_per_job = ENV["FORK_PER_JOB"] != 'false' && Kernel.respond_to?(:fork)
     end
 
     # Returns a symbol representing the current worker state,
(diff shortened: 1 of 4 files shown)


fw42:
ping @steveklabnik

lib/resque/worker.rb
```diff
@@ -226,33 +227,34 @@ def work(interval = 5.0, &block)
...
+          if fork_per_job?
+            run_hook :before_fork, job
             begin
-              Process.waitpid(@child)
-            rescue SystemCallError
-              nil
+              @child = fork do
+                unregister_signal_handlers if term_child
+                perform(job, &block)
+                exit! unless run_at_exit_hooks
```
fw42: This used to say `run_at_exit_hooks ? exit : exit!`. Where did the `exit` (without the `!` go?). Is that unnecessary now because the child exits automatically when the `fork` block ends?

lib/resque/worker.rb
```diff
@@ -226,33 +227,34 @@ def work(interval = 5.0, &block)
...
+              begin
+                Process.waitpid(@child)
+              rescue SystemCallError
+                nil
+              end
+              job.fail(DirtyExit.new("Child process received unhandled signal #{$?.stopsig}")) if $?.signaled?
+              @child = nil
             end
           end
+          unless fork_per_job?
```
fw42: you have a `if fork_per_job?` and then an `unless fork_per_job?`. Couldn't this just be an `else` instead?

fw42: oh, you need to check the condition again because `@fork_per_job` might have changed? Does any test break if I change this to `else`?

test/worker_test.rb
```diff
@@ -891,15 +891,11 @@ def self.perform
...
-        ENV["FORK_PER_JOB"] = 'false'
-        assert !workerA.will_fork?
-      ensure
-        ENV["FORK_PER_JOB"] = 'true'
-      end
+    begin
+      ENV["FORK_PER_JOB"] = 'false'
+      assert_equal false, Resque::Worker.new(:jobs).fork_per_job?
+    ensure
+      ENV["FORK_PER_JOB"] = 'true'
```
fw42: nitpick: Shouldn't we reset this to the previous value (rather than resetting it to true)?

test/worker_test.rb
```diff
@@ -891,15 +891,11 @@ def self.perform
...
-    if workerA.will_fork?
-      begin
-        ENV["FORK_PER_JOB"] = 'false'
-        assert !workerA.will_fork?
-      ensure
-        ENV["FORK_PER_JOB"] = 'true'
-      end
+    begin
+      ENV["FORK_PER_JOB"] = 'false'
+      assert_equal false, Resque::Worker.new(:jobs).fork_per_job?
```
fw42: Why not like this
```ruby
it "won't fork if ENV['FORK_PER_JOB'] is false" do
  without_forking do
    assert_equal false, Resque::Worker.new(:jobs).fork_per_job?
  end
end
```
Isn't that the same thing?

fw42:
This looks much cleaner to me. Thanks Dylan!

fw42:
@jpittis would you mind reviewing this as well please?

lib/resque/worker.rb
dylanahsmith: Yup, the child exits at the end of the fork block.  This is just to maintain the behaviour of skipping the at exit hooks

lib/resque/worker.rb
dylanahsmith: Yup, the extra check is for the fork fallback.  It looks like we are missing a test for that, since changing to an `else` doesn't cause any tests to fail.

test/worker_test.rb
dylanahsmith: Sure.  Pushed a fix for that.

test/worker_test.rb
dylanahsmith: Since the test is explicitly testing the environment variable, I thought modifying the environment variable directly would be clearer

lib/resque/worker.rb
dylanahsmith: Added a test that will fail if this separate `unless fork_per_job?` condition gets changed to an `else`

test/worker_test.rb
```diff
@@ -982,6 +980,31 @@ def self.perform
...
+    end
+    self.perform_count = 0
+
+    def self.perform
+      self.perform_count += 1
+    end
+  end
+
+  it "runs jobs without forking if fork isn't implemented" do
+    nil while @worker.reserve # empty queue
```
fw42: why would the queue not be empty already? isn't redis flushed after each test?

fw42:
This looks good to me. 👍 

Should we try this out in production first before merging? Or at least try and see if the Shopify tests pass with this branch?

@Sirupsen or @steveklabnik, if one of you could review this, that would be great too (since I don't have commit access).

test/worker_test.rb
dylanahsmith: Even if it is, the `before` block enqueues a job https://github.com/dylanahsmith/resque/blob/350d1b4306d0cc74aebc904e5d0e1e0a15a849d2/test/worker_test.rb#L22-L25

fw42:
re-ping, @Sirupsen, @jpittis, @steveklabnik, if someone else could review this too, that would be ❤️

jpittis:
I can't find anything wrong with this. It seems to be making the work loop much cleaner which is awesome. Don't have enough context to +1.

Sirupsen:
> When Kernel.fork is used without a block, then there will be a race condition where an exception can occur from a signal that causes the child process to leave its code path and execute exceptions handlers for the parent process.

How did you discover this?

How does passing it a block change this? Does it ensure the child code path is run before? Isn't there still a race where the child can get a signal before `unregister_signal_handlers` is called from the child?

dylanahsmith:
@Sirupsen we used to get uncaught exceptions from that race condition before we disabled forking in resque.  If we use fork with a block, then ruby ensures that the child exits if it gets an exception in the block given for fork.  Trying to do that without a block means that we will have an exception handler that will have to handle exceptions from the parent or child, which I know was a problem in the past.

> Does it ensure the child code path is run before?

Nope, that isn't necessary.

> Isn't there still a race where the child can get a signal before `unregister_signal_handlers` is called from the child?

Yes.  I don't know if that has caused us any problems though.  The signals mostly just set an instance variable or do nothing if `@child` isn't set.  If we did want to avoid a race condition there, then we could check if `Process.pid` equals Resque::Worker#pid.

Sirupsen:
> If we did want to avoid a race condition there, then we could check if Process.pid equals Resque::Worker#pid.

Do you think it's OK to skip that? I don't have a strong opinion on this.

dylanahsmith:
I think fixing the signal handling race condition should be done in separate PR.

Sirupsen:
LGTM

Sirupsen:
@chrisccerami got time to look at this? ✨

dylanahsmith:
This has been running in production at Shopify, and isn't getting any more feedback, so I'm going to merge it.

jeremy:
Note for plugin authors: if you twiddle `@cant_fork` directly, like https://github.com/stulentsev/resque-multi-job-forks, then you'll need to switch to setting `worker.fork_per_job = false` instead.

PR #1471, opened 2016-06-14
Fix passing an Redis::Namespace object into Resque.redis=

dylanahsmith:
@chrisccerami & @fw42 for review

## Problem

`@data_store` on Resque was being set to a Redis::Namespace instance instead of a Resque::DataStore instance if Resque.redis= was given a Redis::Namespace instance.

## Solution

* Wrap the Resque::Namespace instance with Resque::DataStore like all the other assignments
* Fix the test that expected this broken behaviour

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -80,7 +80,7 @@ module Resque
     # is O(N) for the keyspace, so be careful - this can be slow for big databases.
     def all_resque_keys
       @redis.keys("*").map do |key|
-        key.sub("#{Resque.redis.namespace}:", '')
+        key.sub("#{@redis.namespace}:", '')
       end
     end
 
--- a/test/resque_test.rb
+++ b/test/resque_test.rb
@@ -27,9 +27,9 @@ describe "Resque" do
     new_redis = Redis.new(:host => "localhost", :port => 9736)
     new_namespace = Redis::Namespace.new("namespace", :redis => new_redis)
     Resque.redis = new_namespace
-    assert_equal new_namespace, Resque.redis
 
-    Resque.redis = 'localhost:9736/namespace'
+    assert_equal new_namespace.client, Resque.redis.client
+    assert_equal 0, Resque.size(:default)
   end
 
   it "can put jobs on a queue" do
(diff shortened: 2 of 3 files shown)


lib/resque/data_store.rb
```diff
@@ -80,7 +80,7 @@ def reconnect
     # is O(N) for the keyspace, so be careful - this can be slow for big databases.
     def all_resque_keys
       @redis.keys("*").map do |key|
-        key.sub("#{Resque.redis.namespace}:", '')
```
dylanahsmith: This was only working since Resque.redis normally returns this instance and then namespace goes through method_missing.

fw42:
@davetron5000, mind taking a look?

test/resque_test.rb
```diff
@@ -27,9 +27,9 @@
     new_redis = Redis.new(:host => "localhost", :port => 9736)
     new_namespace = Redis::Namespace.new("namespace", :redis => new_redis)
     Resque.redis = new_namespace
-    assert_equal new_namespace, Resque.redis
 
-    Resque.redis = 'localhost:9736/namespace'
```
fw42: where did this go?

fw42: ah I guess teardown takes care of this

fw42:
Makes sense to me 👍

davetron5000:
This looks good to me!

davetron5000:
Nice catch @dylanahsmith   I feel this implies a missing test somewhere that should've blown up on my original change, but it's not clear to me where that test should've been :(

test/resque_test.rb
dylanahsmith: I think this was added to test this case, but we should only test one thing per test and there is a "can set a namespace through a url-like string" test above that tests this case

dylanahsmith:
@davetron5000 the modified test `it "redis= works correctly with a Redis::Namespace param" do` should now properly cover this case.  In addition to checking the the redis client is the same, it also uses the data store through `Resque.size(:default)` which would have failed without this fix

davetron5000:
@dylanahsmith awesome.  thanks!

dylanahsmith:
@fw42 :shipit: ?  I don't have commit access, so you will need to do it for me

Sirupsen:
Merged to master and cherry-picked to 1-x-stable

PR #1472, opened 2016-06-14
Fix auto-connect to redis when Resque.redis isn't set

dylanahsmith:
@fw42 & @chrisccerami for review

## Problem

On v1.26.0 and this branch (i.e. the expected behaviour)

```
$ bundle exec irb -Ilib -rresque
irb(main):001:0> Resque.size(:default)
=> 0
```

but on master I get

```
NoMethodError: undefined method `queue_size' for nil:NilClass
```

because Resque.redis was auto-connecting when it hadn't been set, but now the Resque module uses `@data_store` which loses that auto-connect behaviour.

## Solution

Use `data_store` instead of `@data_store` since the data_store method auto-connects to redis when Resque.redis isn't set.

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -143,7 +143,7 @@ module Resque
   alias :data_store :redis
 
   def redis_id
-    @data_store.identifier
+    data_store.identifier
   end
 
   # Set or retrieve the current logger object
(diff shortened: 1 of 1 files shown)


fw42:
test?

dylanahsmith:
I'm not really sure how to test this, other than using `Resque.instance_variable_set(:@data_store, nil)`

Any suggestions?

fw42:
Code LGTM. @Sirupsen?

Sirupsen:
LGTM

Issue #1473, opened 2016-06-15
1-x-stable / master


Sirupsen:
@steveklabnik @chrisccerami @davetron5000 do any of you know what the deal is with these two branches? I merged two PRs to master, and ended up cherry-picking them to 1-x-stable. Is 1-x-stable only for released tags? Can we get rid of 1-x-stable now that master is no longer Resque 2?

Sirupsen:
cc @dylanahsmith @fw42

steveklabnik:
`1-x-stable` should be vestigial at this point. I didn't delete it because there were some open PRs I didn't want to force contributors to move over.

steveklabnik:
(so yes, you should be able to just get rid of it onw)

Sirupsen:
Thanks @steveklabnik. I'll delete the branch and  👀  at current PRs against the now defunct branch.

Great work cleaning up all those old PRs ✨

Sirupsen:
Argh, there's a larger difference between the two branches than I expected :/ I have to run, but maybe @dylanahsmith can take a look at this (since you're looking at Resque right now).

steveklabnik:
At the time I left it, there wasn't :sob:

dylanahsmith:
There are 6 merges into 1-x-stable since v1.26.0.

* #1440 merged 1-x-stable to bring in #1439 and #1425
* **#1403 was merged into 1-x-stable and is missing in master**
* #1210 cherry-picked #1466 from 1-x-stable as commit 4bb44413a045fca78d25e22ad535b3f5ea97fab8 without a reference to the PR number
* #1472 and #1471 got cherry-picked from master as squashed merge commits

@chrisccerami you merged #1403 into 1-x-stable, would you mind bringing that into master.  Please keep a reference to the original PR number in the commit you bring over

fw42:
Any objections to deleting the 1-x-stable branch now?

dylanahsmith:
No objections from me

PR #1474, opened 2016-06-20
Cherry-pick 1-x-stable changes into master

fw42:
This cherry-picks https://github.com/resque/resque/pull/1403 / https://github.com/resque/resque/commit/1deabd94d5091db38e05e9ad838d5066126aa1ea (from the 1-x-stable branch) into the master branch, in order to make sure that all changes that are on 1-x-stable are also on master. 

After this is merged, I think we can delete the 1-x-stable branch and just use master.

Closes https://github.com/resque/resque/issues/1473.

@dylanahsmith @Sirupsen

### DIFF
--- a/lib/resque/failure/redis.rb
+++ b/lib/resque/failure/redis.rb
@@ -48,6 +48,10 @@ module Resque
       end
 
       def self.each(offset = 0, limit = self.count, queue = :failed, class_name = nil, order = 'desc')
+        if class_name
+          original_limit = limit
+          limit = count
+        end
         all_items = limit == 1 ? [all(offset,limit,queue)] : Array(all(offset, limit, queue))
         reversed = false
         if order.eql? 'desc'
@@ -55,7 +59,7 @@ module Resque
           reversed = true
         end
         all_items.each_with_index do |item, i|
-          if !class_name || (item['payload'] && item['payload']['class'] == class_name)
+          if !class_name || (item['payload'] && item['payload']['class'] == class_name && (original_limit -= 1) >= 0)
             if reversed
               id = (all_items.length - 1) + (offset - i)
             else
--- a/lib/resque/failure/base.rb
+++ b/lib/resque/failure/base.rb
@@ -53,15 +53,15 @@ module Resque
       # A URL where someone can go to view failures.
       def self.url
       end
-      
+
       # Clear all failure objects
       def self.clear(*args)
       end
-      
-      def self.requeue(index)
+
+      def self.requeue(*args)
       end
 
-      def self.remove(index)
+      def self.remove(*args)
       end
 
       # Logging!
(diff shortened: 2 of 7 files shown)


lib/resque/failure/redis.rb
```diff
@@ -48,14 +48,18 @@ def self.all(offset = 0, limit = 1, queue = nil)
       end
 
       def self.each(offset = 0, limit = self.count, queue = :failed, class_name = nil, order = 'desc')
+        if class_name
+          original_limit = limit
+          limit = count
```
dylanahsmith: These variables aren't very descriptive of there purpose.  Especially when `original_limit` is decremented (original to me implies that it is read-only).

I would prefer `fetch_limit = class_name ? count : limit` where the following line count use `fetch_limit` instead.

fw42: legit concern, but I'm just cherry-picking this (it has already been merged). I think if we have concerns about this code, it should be addressed outside of this PR (I can do that if you like, but not here I think).

lib/resque/failure/redis.rb
```diff
@@ -48,14 +48,18 @@ def self.all(offset = 0, limit = 1, queue = nil)
...
+        end
         all_items = limit == 1 ? [all(offset,limit,queue)] : Array(all(offset, limit, queue))
```
dylanahsmith: off topic, but
* `Array(` doesn't do anything when `all` returns an array
* `all` should just always return an array and there should be another method to get a single item from the queue, so this can just be `all_items = all(offset, limit, queue)`

lib/resque/failure/redis.rb
```diff
@@ -48,14 +48,18 @@ def self.all(offset = 0, limit = 1, queue = nil)
...
         reversed = false
         if order.eql? 'desc'
           all_items.reverse!
           reversed = true
         end
         all_items.each_with_index do |item, i|
-          if !class_name || (item['payload'] && item['payload']['class'] == class_name)
+          if !class_name || (item['payload'] && item['payload']['class'] == class_name && (original_limit -= 1) >= 0)
```
dylanahsmith: This would be more readable if the added condition was `&& original_limit > 0` and the decrement was moved inside the if block as just `original_limit -= 1`.  It will also be more readable with the `original_limit` -> `limit` rename as suggested above

lib/resque/failure/redis.rb
dylanahsmith: Why do we even get the count?  It seems like it should be possible to use a `nil` limit (i.e. `-1` limit in the redis lrange command)

dylanahsmith:
I guess this PR was already reviewed, so just get CI 🍏

dylanahsmith:
It was just a flaky test failure, so :+1:

fw42:
I don't have commit access.. can someone merge?

PR #1478, opened 2016-06-22
Kill all heartbeat threads after each test run

fw42:
See https://github.com/resque/resque/pull/1477. Prevent threads from leaking across tests.

@dylanahsmith @Sirupsen

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -16,6 +16,12 @@ module Resque
 
     WORKER_HEARTBEAT_KEY = "workers:heartbeat"
 
+    @@all_heartbeat_threads = []
+    def self.kill_all_heartbeat_threads
+      @@all_heartbeat_threads.each(&:kill)
+      @@all_heartbeat_threads = []
+    end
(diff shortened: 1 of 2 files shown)


lib/resque/worker.rb
```diff
@@ -16,6 +16,12 @@ class Worker
 
     WORKER_HEARTBEAT_KEY = "workers:heartbeat"
 
+    @@all_heartbeat_threads = []
```
dylanahsmith: Couldn't you just use `@all_heartbeat_threads` since it will be an instance variable on the class?  Or are you trying to make this kill threads for subclasses if there happen to be any?

dylanahsmith: Oh, I guess using `@@` means that you can reference it from an instance method.

fw42: That would work too, yea. I don't really care either way.

dylanahsmith:
:+1:

Sirupsen:
:+1:

PR #1479, opened 2016-06-27
Always perform heatbeats in the heartbeat thread

fw42:
I think it's cleaner to always perform the heartbeats in the actual heartbeat thread, never outside of it (and this makes another refactor that I'm working on a little bit easier).

@dylanahsmith @Sirupsen @jpittis

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -503,17 +503,15 @@ module Resque
     end
 
     def start_heartbeat
-      heartbeat!
-
       @stop_heartbeat_thread = false
       @heart = Thread.new do
         until @stop_heartbeat_thread do
+          heartbeat!
+
           Resque.heartbeat_interval.times do
             sleep(1)
             break if @stop_heartbeat_thread
           end
-
-          heartbeat!
         end
       end
 


Sirupsen:
LGTM

lib/resque/worker.rb
```diff
@@ -503,17 +503,15 @@ def heartbeat!(time = Time.now)
     end
 
     def start_heartbeat
-      heartbeat!
```
jpittis: Is there a specific reason this was being done outside the thread to begin with?

Sirupsen: I don't remember there being one, and I wrote it. 🙈  Pretty sure I would've written a failing test if there was.

jpittis:
LGTM besides comment. Since @Sirupsen added that code and he signs off, I can't see there being a problem.

lib/resque/worker.rb
dylanahsmith: I think this should be done outside the thread so that it is done synchronously resque starts processing jobs.  Otherwise, there is a race condition where the first job could start without a up to date heartbeat value written to redis. Unfortunately, race conditions aren't easy to write a failing test for.

fw42: > race condition where the first job could start without a up to date heartbeat value

why is that a problem?

dylanahsmith: because pruning a worker that is still running a job can cause it to be re-enqueued for a retry while still running. So the job could be running twice concurrently

fw42: As far as I understand it, we aren't pruning if you don't have a heartbeat, we only prune if there is a heartbeat but it expired (i.e. the key exists but the timestamp is too old).

dylanahsmith: oh ok

Issue #1357, opened 2015-12-03
Multiple workers on same queue


mojovski:
When I start multiple workers for the same queue, none of them is displayed in the resque console (admin view).
I dont have an error message or something, so would be great if you can advise me how I can support you best at this issue.

Thanks!

RochesterinNYC:
When you pull up the Rails console (are you using Rails or no?) and run `Resque.workers`, do you see any workers ? Wondering if your workers are actually running and it's a resque-web related problem or if your workers silently fail or something.

Any other details about how you're starting the workers, queue setup, etc. would help as well.

mojovski:
ok, got new insight.
I have two workers on different server.
Both are processing something. I can see this from the htop console via 
``` pstree -cpa ```
In the terminal,
only one worker is set to working:
```
Resque.workers.map { |w| w if w.working? }.compact
```
The worker, which is not visible as working was started via
```
RAILS_ENV=production BACKGROUND=yes QUEUE="${w}" rake resque:work
```
with w as the worker name.
The visible worker was started on another machine with
```
QUEUE="${w}" rake resque:work
```

I will keep investigating it, but if you have an idea where to look, I would really appreciate it. :)

thank you!

**Comment by steveklabnik** [comment]
It would be nice to get a reproduction of this issue.

**Comment by mojovski** [comment]
I will probably find some time for this next week. Sorry for the delay

**Comment by pjammer** [comment]
I can replicate this right now. I was on the last tagged Resque, `1.25.2 `and using capistrano-resque. I just switched to this `1-x-stable` branch and now the resque web console shows the # of workers that Resque.workers has in the `rails console` but that does not match up with the available pids.  i set 3 worker queues, with 2 workers each. There are 6 pids created, but in rails console, Resque.workers shows 4, and only 1 of them matches a pid number in ps aux, yet ps aux has 6 workers running and waiting for my queues.
### capistrano resque runs startup command  during a deploy (this is .6, but there are 5 more as well)
```bash
/usr/bin/env nohup bundle exec rake RAILS_ENV=production QUEUE="low" PIDFILE=/home/deploy/fakeapp/shared/tmp/pids/resque_work_6.pid BACKGROUND=yes VERBOSE=1 INTERVAL=5 environment resque:work >> /home/deploy/fakeapp/shared/log/resque.log 2>> /home/deploy/fakeapp/shared/log/resque.log
```
### rails console
```ruby
irb(main):001:0> Resque.workers
[...]

PR #1480, opened 2016-06-30
[Bugfix] Heartbeat with server time, not worker time

fw42:
This addresses an issue mentioned in https://github.com/resque/resque/issues/1357#issuecomment-229613020 where a worker might prune previous workers if it's clock has significant drift compared to the other workers.

This PR attempts to fix that by using the server time for all heartbeat timestamps instead of the local worker's clock.

@Sirupsen @dylanahsmith @hackhowtofaq

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -484,13 +484,14 @@ module Resque
     def self.all_workers_with_expired_heartbeats
       workers = Worker.all
       heartbeats = Worker.all_heartbeats
+      now = server_time
 
       workers.select do |worker|
         id = worker.to_s
         heartbeat = heartbeats[id]
 
         if heartbeat
-          seconds_since_heartbeat = (Time.now - Time.parse(heartbeat)).to_i
+          seconds_since_heartbeat = (now - Time.parse(heartbeat)).to_i
           seconds_since_heartbeat > Resque.prune_interval
         else
           false
@@ -498,7 +499,12 @@ module Resque
       end
     end
 
-    def heartbeat!(time = Time.now)
+    def self.server_time
+      time, _ = data_store.time
+      Time.at(time)
+    end
+
+    def heartbeat!(time = self.class.server_time)
       redis.hset(WORKER_HEARTBEAT_KEY, to_s, time.iso8601)
     end
 
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -660,12 +660,32 @@ describe "Resque::Worker" do
     assert_equal 1, Resque.workers.size
   end
 
-  it "does return returns a valid time when asking for heartbeat" do
+  it "does return a valid time when asking for heartbeat" do
     workerA = Resque::Worker.new(:jobs)
     workerA.register_worker
     workerA.heartbeat!
 
     assert_instance_of Time, workerA.heartbeat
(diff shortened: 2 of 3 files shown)


lib/resque/worker.rb
```diff
@@ -484,21 +484,27 @@ def self.all_heartbeats
...
           seconds_since_heartbeat > Resque.prune_interval
         else
           false
         end
       end
     end
 
-    def heartbeat!(time = Time.now)
+    def self.server_time
+      time, _ = redis.time
```
dylanahsmith: I think access to redis is supposed to be done through `data_store` now, with Resque::DataStore delegating to redis

test/worker_test.rb
```diff
@@ -660,14 +660,35 @@ def self.perform
...
     assert_instance_of Time, workerA.heartbeat
   end
 
+  it "does not generate heartbeats that depend on the worker clock, but only on the server clock" do
+    server_time = Resque::Worker.server_time
+    fake_time = Time.parse("2000-01-01")
+
+    with_fake_time(fake_time) do
+      worker_time = Time.now
+      assert_equal fake_time, worker_time
```
dylanahsmith: Isn't this assertion really a test for with_fake_time?  Should we test our test helpers like this in a separate test?

test/worker_test.rb
```diff
@@ -660,14 +660,35 @@ def self.perform
...
+
+      workerA = Resque::Worker.new(:jobs)
+      workerA.register_worker
+      workerA.heartbeat!
+
+      heartbeat_time = workerA.heartbeat
+      refute_equal heartbeat_time, worker_time
+
+      delta = 5
```
dylanahsmith: Couldn't you just call `Resque::Worker.server_time` again to get the time range that it should fall under more precisely?

test/worker_test.rb
fw42: yea you're right, will remove this here

test/worker_test.rb
fw42: good idea

lib/resque/worker.rb
fw42: I think `redis` here is actually a DataStore instance, so that should all "just work"

Sirupsen:
LGTM aside from Dylan's comments!

fw42:
updated the test to not use this `delta` hack

lib/resque/worker.rb
dylanahsmith: To me this usage seems like it is deprecated based on the comment in DataStore#method_missing (https://github.com/resque/resque/blob/160ac190e94acead7289e9c815d1ef8b0a15843c/lib/resque/data_store.rb#L50-L54).  Also, using `redis` like this seems like it defeats the purpose of https://github.com/resque/resque/commit/4bb44413a045fca78d25e22ad535b3f5ea97fab8.  If we are going to put an abstraction between resque and redis, then this is a leaky abstraction.

method_missing is also slower to use than a normal method call.

fw42: you are right. I changed it. Thanks!

Sirupsen:
LGTM

dylanahsmith:
:+1:

PR #1481, opened 2016-06-30
Run rails before_eager_load hooks in resque setup

abhi-patel:
Rails has added support for running hooks before eager load. Resque should also be executing the same hooks before calling `Rails.application.eager_load!`

### DIFF
--- a/lib/resque/tasks.rb
+++ b/lib/resque/tasks.rb
@@ -39,13 +39,19 @@ namespace :resque do
 
   # Preload app files if this is Rails
   task :preload => :setup do
-    if defined?(Rails) && Rails.respond_to?(:application)
-      # Rails 3
-      Rails.application.eager_load!
-    elsif defined?(Rails::Initializer)
-      # Rails 2.3
-      $rails_rake_task = false
-      Rails::Initializer.run :load_application_classes
+    if defined?(Rails)
+      if Rails::VERSION::MAJOR > 3
+        ActiveSupport.run_load_hooks(:before_eager_load, Rails.application)
+        Rails.application.config.eager_load_namespaces.each(&:eager_load!)
+
+      elsif Rails::VERSION::MAJOR == 3
+        ActiveSupport.run_load_hooks(:before_eager_load, Rails.application)
+        Rails.application.eager_load!
+
+      elsif defined?(Rails::Initializer)
+        $rails_rake_task = false
+        Rails::Initializer.run :load_application_classes
+      end
     end
   end
 


abhi-patel:
Checkout https://github.com/rails/rails/blob/master/railties/lib/rails/application/finisher.rb#L58

Rails explicitly runs before_eager_load hooks while eager loading. Rescue is missing the same and hence we ran into some issues where we tried running some code when eager load is enabled. 

Everything worked fine for normal code while the same code was throwing exceptions for Resque workers.

aws/aws-sdk-ruby#1229

fw42:
@rafaelfranca, does this make sense to you?

rafaelfranca:
Yes it does. But this code is not eager loading engines for example so this kind of issues can still happen when using an engine. I'll comment inline what can be done but it will not work for Rails 3.

lib/resque/tasks.rb
```diff
@@ -41,6 +41,7 @@
   task :preload => :setup do
     if defined?(Rails) && Rails.respond_to?(:application)
       # Rails 3
+      ActiveSupport.run_load_hooks(:before_eager_load, Rails.application)
       Rails.application.eager_load!
```
rafaelfranca: For Rails 4 it should be `Rails.application.config.eager_load_namespaces.each(&:eager_load!)`

abhi-patel:
@fw42 - Would It be fine to add condition here based on Rails version? Or check if it responds to 'eager_load_namespaces' method?

lib/resque/tasks.rb
abhi-patel: `Rails.application.config.eager_load_namespaces.each(&:eager_load!)` will run eager_load for specified classes/engines.

I suppose will still have to call `ActiveSupport.run_load_hooks(:before_eager_load, Rails.application)` to run the hooks, Or will they execute automatically(internally) by above line(eager_load!)?

rafaelfranca: You are still supposed to call that line to run the hooks.

abhi-patel:
Look like builds failed because of gemfile isssue. Should be fine now.

abhi-patel:
@fw42 - Any update on this?

fw42:
Can you rebase your branch? I think those CI failures have been fixed on master.

As for the actual changes: I don't feel qualified to review this, but if @rafaelfranca thinks that the changes here make sense and if one other person from resque can +1 this, then I'm happy to merge.

@steveklabnik?

steveklabnik:
So, this looks fine to me, but I will admit that I'm a bit out of the loop on exactly what's required to run when here. If @rafaelfranca says :+1: I'm :+1:

abhi-patel:
Just Rebased. Its still broken because of some other gem issue.

fw42:
@rafaelfranca does this look good to merge to you?

rafaelfranca:
:shipit:

PR #1484, opened 2016-07-03
Wait for heartbeat threads to die

fw42:
follow-up to https://github.com/resque/resque/pull/1478

turns out we have to wait for the threads today, otherwise stuff seems to still be leaking.

also stub signal handler registration and unregistration, since those would also leak into other tests.

@dylanahsmith @Sirupsen

### DIFF
--- a/lib/resque/thread_signal.rb
+++ b/lib/resque/thread_signal.rb
@@ -1,0 +1,45 @@
+class Resque::ThreadSignal
+  if RUBY_VERSION <= "1.9"
+    def initialize
+      @signaled = false
+    end
+
+    def signal
+      @signaled = true
+    end
+
+    def wait_for_signal(timeout)
+      (10 * timeout).times do
+        sleep(0.1)
+        return true if @signaled
+      end
+
+      @signaled
+    end
+
+  else
+    def initialize
+      @mutex = Mutex.new
+      @signaled = false
+      @received = ConditionVariable.new
+    end
+
+    def signal
+      @mutex.synchronize do
+        @signaled = true
+        @received.signal
+      end
+    end
+
+    def wait_for_signal(timeout)
+      @mutex.synchronize do
+        unless @signaled
+          @received.wait(@mutex, timeout)
+        end
+
+        @signaled
+      end
+    end
+
+  end
+end
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -18,7 +18,7 @@ module Resque
 
     @@all_heartbeat_threads = []
     def self.kill_all_heartbeat_threads
-      @@all_heartbeat_threads.each(&:kill)
+      @@all_heartbeat_threads.each(&:kill).each(&:join)
       @@all_heartbeat_threads = []
     end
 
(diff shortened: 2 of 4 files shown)


test/worker_test.rb
```diff
@@ -21,13 +21,11 @@ def messages
 
   before do
     @worker = Resque::Worker.new(:jobs)
+    Resque::Worker.any_instance.stubs(:register_signal_handlers)
+    Resque::Worker.any_instance.stubs(:unregister_signal_handlers)
     Resque::Job.create(:jobs, SomeJob, 20, '/tmp')
   end
 
-  after do
-    @worker.kill_background_threads
```
fw42: this is unnecessary now, since https://github.com/resque/resque/pull/1478 takes care of it

test/test_helper.rb
```diff
@@ -56,6 +56,7 @@ def before_setup
 
   def after_teardown
     Resque::Worker.kill_all_heartbeat_threads
+    sleep 0.1 until Thread.list.count == 1
```
dylanahsmith: we might need to worry about a similar leak in production.  We should join on the threads to make sure they die. That is also needed to properly unregister the worker and cleanup its heartbeat

lib/resque/worker.rb
```diff
@@ -18,7 +18,7 @@ class Worker
 
     @@all_heartbeat_threads = []
     def self.kill_all_heartbeat_threads
-      @@all_heartbeat_threads.each(&:kill)
+      @@all_heartbeat_threads.each(&:kill).each(&:join)
```
fw42: @dylanahsmith: like this?

fw42:
@dylanahsmith: Updated. Not sure if I'm doing this right, might be a bit overkill, but I wanted to get rid of the ugly `sleep` in the loop (otherwise the `join` in `kill_background_threads` would always wait for at least a second, and therefore make the tests pretty slow). I think (!) this seems cleaner now? sleeping for thread coordination seems like a bad idea anyway...

lib/resque/thread_signal.rb
```diff
@@ -0,0 +1,24 @@
+class Resque::ThreadSignal
```
chrisccerami: I like this abstraction a lot. 👍

fw42:
Hm looks like this doesn't seem to work on Ruby 1.8.x. Is that still officially supported?

chrisccerami:
As far as i know, yes.

fw42:
@steveklabnik: Do we care about Ruby 1.8?

lib/resque/worker.rb
```diff
@@ -509,19 +509,17 @@ def heartbeat!(time = self.class.server_time)
...
+      @heartbeat_thread_signal = Resque::ThreadSignal.new
 
-          Resque.heartbeat_interval.times do
-            sleep(1)
-            break if @stop_heartbeat_thread
-          end
+      @heartbeat_thread = Thread.new do
+        loop do
+          heartbeat!
+          signaled = @heartbeat_thread_signal.wait_for_signal(Resque.heartbeat_interval)
```
Sirupsen: 👍

Sirupsen:
I like this 👍

dylanahsmith:
:+1:

chrisccerami:
I feel like dropping 1.8 support constitutes a breaking change.

dylanahsmith:
On ruby 1.8.7 ConditionalVariable doesn't provide a way to wait with a timeout.  We could sleep for the timeout instead on 1.8.7, but that would make the tests very slow in ruby 1.8.7.

Would it be safe to Thread#kill the heartbeat thread instead?  That would work on ruby 1.8.7.

fw42:
@chrisccerami, I added Ruby 1.8 compatible code. It's shitty, but not any more shitty than what is on master right now. Does this work for you?

dylanahsmith:
:+1:

PR #1485, opened 2016-07-03
[bugfix] Remove heartbeat when unregistering worker

fw42:
This fixes a bug that was introduced in https://github.com/resque/resque/commit/4bb44413a045fca78d25e22ad535b3f5ea97fab8 where old heartbeats would no longer get cleaned up when a worker is being shut down. By itself, this is not a big deal, but in combination with https://github.com/resque/resque/pull/1479, it can lead to a race condition where workers that re-use previous worker ids (because hostname+pid+queues happen to match those of a previous worker) get pruned.

Also move all the heartbeat methods to the DataStore class.

@dylanahsmith @Sirupsen

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -4,6 +4,8 @@ module Resque
   class DataStore
     extend Forwardable
 
+    HEARTBEAT_KEY = "workers:heartbeat"
+
     def initialize(redis)
       @redis               = redis
       @queue_access        = QueueAccess.new(@redis)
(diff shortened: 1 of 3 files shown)


Sirupsen:
LGTM

dylanahsmith:
:+1:

lib/resque/data_store.rb
```diff
@@ -84,6 +89,11 @@ def all_resque_keys
       end
     end
 
+    def server_time
+      time, _ = @redis.time
```
davegudge: It looks like this change is generating the following deprecation warning:

> Passing 'time' command to redis as is; blind passthrough has been deprecated and will be removed in redis-namespace 2.0

davegudge: I see the issue will be resolved by updating `redis-namespace`

https://github.com/resque/redis-namespace/pull/119

PR #1486, opened 2016-07-05
Add more tests

fw42:
Follow-up to https://github.com/resque/resque/pull/1485.

Just add some more tests for this to ensure workers are also cleaned up when `prune_dead_workers` is being called (this is implicitly true because `prune_dead_workers` calls `unregister_worker`, but I feel like it deserves dedicated test coverage).

@Sirupsen @dylanahsmith if one of you could take a quick look?

### DIFF
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -627,33 +627,41 @@ describe "Resque::Worker" do
   end
 
   it "prunes dead workers with heartbeat older than prune interval" do
+    assert_equal({}, Resque::Worker.all_heartbeats)
     now = Time.now
 
     workerA = Resque::Worker.new(:jobs)
-    workerA.instance_variable_set(:@to_s, "bar:3:jobs")
+    workerA.to_s = "bar:3:jobs"
     workerA.register_worker
     workerA.heartbeat!(now - Resque.prune_interval - 1)
 
     assert_equal 1, Resque.workers.size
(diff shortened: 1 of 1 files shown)


test/worker_test.rb
```diff
@@ -654,10 +661,12 @@ def self.perform
     workerA.register_worker
 
     assert_equal 1, Resque.workers.size
+    assert_equal({}, Resque::Worker.all_heartbeats)
 
     @worker.prune_dead_workers
 
     assert_equal 1, Resque.workers.size
+    assert_equal({}, Resque::Worker.all_heartbeats)
```
dylanahsmith: Why bother with this assertion?  There wasn't any heartbeat to remove in this case.  Are you worried that prune_dead_workers will set a heartbeat?

Sirupsen:
👍

test/worker_test.rb
```diff
@@ -635,17 +636,23 @@ def self.perform
     workerA.heartbeat!(now - Resque.prune_interval - 1)
 
     assert_equal 1, Resque.workers.size
+    assert Resque::Worker.all_heartbeats.key?(workerA.to_s)
 
     workerB = Resque::Worker.new(:jobs)
     workerB.instance_variable_set(:@to_s, "foo:5:jobs")
```
fw42: updated

PR #1488, opened 2016-07-07
Remove old heartbeats before starting heartbeat thread

fw42:
It seems like https://github.com/resque/resque/pull/1485 did not fix the issue fully. I'm still seeing some PruneDeadWorker exceptions in production, which leads me to believe that there is still a race condition. Sadly I'm not really sure what the race condition is though :-(

This PR should once and for all fix it, by explicitly removing old heartbeats (of the same worker id) before sending any new heartbeats (and before registering the worker).

@dylanahsmith @Sirupsen

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -42,6 +42,7 @@ module Resque
                               :unregister_worker,
                               :heartbeat,
                               :heartbeat!,
+                              :remove_heartbeat,
                               :all_heartbeats,
                               :set_worker_payload,
                               :worker_start_time,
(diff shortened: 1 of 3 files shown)


Sirupsen:
:( I don't have any other ideas on the race. This doesn't seem crazy regardless, so I'm ok with it.

dylanahsmith:
:+1:

fw42:
something is weird with travis CI.. same error as the other branch.. unrelated to the changes here

fw42:
confirmed in production that this solves the errors I was seeing

PR #1489, opened 2016-07-08
Use older json version

fw42:
Not sure why but CI is currently failing for Ruby 1.9 and Ruby 1.8 because it's installing json 2.x which requires Ruby 2.x. See if this fixes it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,13 @@ gemspec
 
 gem "activesupport", "~> 3.0"
 gem "i18n"
-gem "json"
+
+if RUBY_VERSION < "2.0"
+  gem "json", '~> 1.8'
+else
+  gem "json"
+end
+
 gem "minitest", "4.7.0"
 gem "mocha", :require => false
 gem "rack-test", "~> 0.5"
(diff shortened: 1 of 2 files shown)


fw42:
🍏

fw42:
@Sirupsen or @dylanahsmith, does this make sense to you?

Gemfile
```diff
@@ -4,7 +4,13 @@ gemspec
 
 gem "activesupport", "~> 3.0"
 gem "i18n"
-gem "json"
+
+if RUBY_VERSION < "2.0"
+  gem "json", '1.8.3'
```
dylanahsmith: At least use `'~> 1.8'` so it isn't tied to this specific version.

Gemfile
```diff
@@ -4,7 +4,13 @@ gemspec
 
 gem "activesupport", "~> 3.0"
 gem "i18n"
-gem "json"
+
+if RUBY_VERSION < "2.0"
```
dylanahsmith: It looks like https://rubygems.org/gems/json says REQUIRED RUBY VERSION: ~> 2.0

I wonder why that was ignored.

dylanahsmith:
:+1:

PR #1494, opened 2016-08-10
Simplify classify method

m1lt0n:
A small simplification of the utility method classify.


fw42:
hrm, looks like CI is broken again (not your fault I think)

PR #1495, opened 2016-08-11
fall back to standard implementation of respond_to? for data store

m1lt0n:
As a good measure respond_to? could return whether the data store responds to a method of itself, too, as it could lead to bugs in code using the data store and relying on the default behaviour of respond_to?

I might contribute some really small things while familiarizing myself with the codebase of the project. Sorry for perhaps the multiple PRs.

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -61,7 +61,7 @@ module Resque
 
     # make use respond like redis
     def respond_to?(method,include_all=false)
-      @redis.respond_to?(method,include_all)
+      @redis.respond_to?(method,include_all) || super
     end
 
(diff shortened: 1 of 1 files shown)


fw42:
I think you just pushed this same commit to https://github.com/resque/resque/pull/1494 as well.. That probably wasn't intended? Can you use separate branches?

m1lt0n:
@fw42 Yes, you're right, sorry for that. Preparing a PR with a separate branch. You can close the other PR.

fw42:
does this fix a particular problem that you ran into? or "just in case"?

m1lt0n:
@fw42 Just in case. If it seems like overengineering, feel free to close it :)

PR #1375, opened 2016-01-13
Support for adding reverse proxy friendly url prefixes

jpang-h4:
Summary
-
In order for resque-web to be put behind a reverse-proxy path
that is not root, resque-web must serve templates with asset and
url links with the reverse proxy url prefix.

The url prefix can be added using '-a' option. e.g:
`./resque-web -F -L -s thin -p 5678 -e development -a '/proxy_sites/resque'`

Tests Ran:
-
Existing tests with additional url_prefix use case

PR #1497, opened 2016-08-11
Fixed reverse proxy prefix for overview and stats

razvanphp:
This fixes PR #1375 since the overview page and stats page were not using the prefix.

I did not add anything, just reverted the extra changes, works good like this for us.

### DIFF
--- a/lib/resque/server.rb
+++ b/lib/resque/server.rb
@@ -42,10 +42,6 @@ module Resque
       end
       alias_method :u, :url_path
 
-      def redirect_url_path(*path_parts)
-        [ path_prefix, path_parts ].join("/").squeeze('/')
-      end
-
       def path_prefix
         request.env['SCRIPT_NAME']
       end
@@ -158,7 +154,7 @@ module Resque
 
     # to make things easier on ourselves
     get "/?" do
-      redirect redirect_url_path(:overview)
+      redirect url_path(:overview)
     end
 
     %w( overview workers ).each do |page|
@@ -219,7 +215,7 @@ module Resque
 
     post "/failed/:queue/requeue/all" do
       Resque::Failure.requeue_queue Resque::Failure.job_queue_name(params[:queue])
-      redirect redirect_url_path("/failed/#{params[:queue]}")
+      redirect url_path("/failed/#{params[:queue]}")
     end
 
     get "/failed/requeue/:index/?" do
@@ -251,7 +247,7 @@ module Resque
     end
 
     get "/stats/?" do
-      redirect redirect_url_path("/stats/resque")
+      redirect url_path("/stats/resque")
     end
 
     get "/stats/:id/?" do


fw42:
@jpang-h4 does this make sense to you?

jpang-h4:
@razvanphp - I'm actually unable to reproduce what you mentioned in #1375.

If you look at the changes made for #1375 - In particular, https://github.com/resque/resque/pull/1375/files#diff-f384147196b88f2c83e38e710f7bf041R46 is exactly the same as the new `redirect_url_path`, which means your change will be adding unnecessary prefixes.

Just so that I can try to help, how are you triggering the resque-web command? Maybe some combination of option flags are tripping up something that I am not aware of?

razvanphp:
I thought the bug is pretty clear even without testing it, `redirect_url_path` is just a copy of original `url_path` BEFORE `url_prefix` was added, so every time the method is used, no prefix is added.

I tested checked my theory when I realised the prefix option does not work behind simple nginx reverse proxy, after changes in this PR it works great.

The unit test added checks the assets links, which work in both cases, but the overview and status menu links don't.

jpang-h4:
@razvanphp 

> I tested checked my theory when I realised the prefix option does not work behind simple nginx reverse proxy, after changes in this PR it works great.

This is probably key to the issue you found. In my setup, the prefix portion is stripped off from the URL before the second hop URL is attempted. Maybe the specific nginx setting that you have isn't doing the stripping off. I'm of opinion to get nginx to strip off the prefix and not make the change here at resque.

razvanphp:
Maybe I did not make myself clear enough, the issue it's not about stripping any prefix off, it's about the lack of prefix, because the duplicated method `redirect_url_path` simply does not include the prefix at all.

And why is it even needed? What is the difference between `redirect_url_path` and `url_path` now?

The issue here is that for example accessing `https://server.com/resque/job2/stats` redirects you to `https://server.com/stats/resque`, without the prefix `/resque/job2/` that I specified in the command line, and this is not done by nginx.

I start resque-web like this: `resque-web -a /resque/job1` and my nginx reverse proxy config:
```
location /resque/job2/ {
    proxy_pass http://10.10.66.199:5678/;
    proxy_next_upstream error timeout http_404 http_502 http_504;
    proxy_set_header Host $host;
    proxy_set_header X-Real-IP $remote_addr;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $scheme;
}
```
so nothing from here can or should strip anything, actually I don't understand how it works in your case without additional logic in the reverse proxy.

Probably other pair of eyes have to look at this one, as you might be biased, since it's your code, no offence.

Tonkpils:
Can't say I'm very familiar with this issue but has there been a resolution to this @razvanphp /cc @cehoffman

jeremy:
This depends on whether the proxy rewrites Location headers to account for the URL change. #1375 was written with the expectation that the proxy adds the URL prefix back to the redirected URL. Check out http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_redirect for how nginx handles this by default and further config.

jeremy:
Thanks @razvanphp! Using the URL prefix in all cases is what most reverse proxies will expect, and be able to cope with.

Issue #1493, opened 2016-08-09
log writing failed. can't be called from trap context


sonots:
On stopping resque:work with Ctrl-C, I get and error as the title.

I added backtrace to ruby's standard error and below is the its stacktrace

```
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/2.2.0/monitor.rb:185:in `lock'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/2.2.0/monitor.rb:185:in `mon_enter'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/2.2.0/monitor.rb:209:in `mon_synchronize'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/2.2.0/logger.rb:595:in `write'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/2.2.0/logger.rb:378:in `add'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/2.2.0/logger.rb:434:in `info'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/resque-1.26.0/lib/resque/logging.rb:8:in `log'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/resque-1.26.0/lib/resque/worker.rb:872:in `log_with_severity'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/resque-1.26.0/lib/resque/worker.rb:428:in `shutdown'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/resque-1.26.0/lib/resque/worker.rb:435:in `shutdown!'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/resque-1.26.0/lib/resque/worker.rb:390:in `block in register_signal_handlers'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/resque-1.26.0/lib/resque/worker.rb:246:in `call'
/Users/seo.naotoshi/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/resque-1.26.0/lib/resque/worker.rb:246:in `sleep'
[...]

PR #1499, opened 2016-08-17
Fix log writing failed. can't be called from trap context (thread version)

sonots:
Fix https://github.com/resque/resque/issues/1493

### DIFF
--- a/lib/resque/signal_handler.rb
+++ b/lib/resque/signal_handler.rb
@@ -1,10 +1,17 @@
 require 'thread'
 
-class Resque::SignalTrap
+class Resque::SignalHandler
   attr_reader :thread
 
   def initialize
     @handlers = {}
+    reopen
+  end
+
+  def reopen
+    @self_read.close rescue nil if @self_read
+    @self_write.close rescue nil if @self_write
+    @thread.kill rescue nil if @thread and @thread.alive?
     @self_read, @self_write = IO.pipe
     @thread = Thread.new(&method(:handle_signal))
   end
--- a/lib/resque/signal_handler.rb
+++ b/lib/resque/signal_handler.rb
@@ -24,7 +24,13 @@ class Resque::SignalHandler
     end
 
     @handlers[sig] = block
-    Kernel.trap(sig) { @self_write.puts(sig) }
+    Kernel.trap(sig) do
+      @self_write.puts(sig)
+      # Ensure main and `@handlers[sig].call` in handle_signal thread does not run concurrently
+      # so that we do not need to care `@handlers[sig].call` is thread-safe
+      # But, note that we can not receive same signal again until `@handlers[sig].call` finishes
+      Thread.stop
+    end
   end
 
   private
@@ -32,7 +38,12 @@ class Resque::SignalHandler
   def handle_signal
     while readable_io = IO.select([@self_read])
       sig = readable_io.first[0].gets.strip
-      @handlers[sig].call
+      until Thread.main.stop? do; end
+      begin
+        @handlers[sig].call
+      ensure
+        Thread.main.wakeup
+      end
     end
   end
 end
--- a/lib/resque/signal_handler.rb
+++ b/lib/resque/signal_handler.rb
@@ -24,13 +24,7 @@ class Resque::SignalHandler
     end
 
     @handlers[sig] = block
-    Kernel.trap(sig) do
-      @self_write.puts(sig)
-      # Ensure main and `@handlers[sig].call` in handle_signal thread does not run concurrently
-      # so that we do not need to care `@handlers[sig].call` is thread-safe
-      # But, note that we can not receive same signal again until `@handlers[sig].call` finishes
-      Thread.stop
-    end
+    Kernel.trap(sig) { @self_write.puts(sig) }
   end
 
   private
@@ -38,12 +32,7 @@ class Resque::SignalHandler
   def handle_signal
     while readable_io = IO.select([@self_read])
       sig = readable_io.first[0].gets.strip
-      until Thread.main.stop? do; end
-      begin
-        @handlers[sig].call
-      ensure
-        Thread.main.wakeup
-      end
+      @handlers[sig].call
     end
   end
 end
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -373,18 +373,19 @@ module Resque
     # USR2: Don't process any new jobs
     # CONT: Start processing jobs again after a USR2
     def register_signal_handlers
-      trap('TERM') { graceful_term ? shutdown : shutdown!  }
-      trap('INT')  { shutdown!  }
+      st = Resque::SignalTrap.new
+      st.trap('TERM') { graceful_term ? shutdown : shutdown!  }
+      st.trap('INT')  { shutdown!  }
 
       begin
-        trap('QUIT') { shutdown   }
+        st.trap('QUIT') { shutdown   }
         if term_child
-          trap('USR1') { new_kill_child }
+          st.trap('USR1') { new_kill_child }
         else
-          trap('USR1') { kill_child }
+          st.trap('USR1') { kill_child }
         end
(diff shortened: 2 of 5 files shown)


sonots:
CI failed, but it is not related with this commit (failed with ruby 1.9.3 and 1.8.7)

fw42:
Hmmm.. I'm not a fan of how this will make your application multi-threaded in possibly unexpected ways. For example, at Shopify, our Logger is actually logging to Kafka, and with this PR, that Kafka logger would run in a seperate thread. Now I have to think about if the logging code is thread-safe, which I wouldn't expect having to think about in an otherwise single-threaded application.

@dylanahsmith, how do you feel about this?

lib/resque/signal_handler.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def trap(sig, &block)
+    # normalize signal names
+    sig = sig.to_s.upcase
+    if sig[0,3] == "SIG"
+      sig = sig[3..-1]
+    end
+
+    @handlers[sig] = block
+    Kernel.trap(sig) { @self_write.puts(sig) }
```
dylanahsmith: Isn't it the logging not the signal handling that needs to be done in another thread?

dylanahsmith:
If you want a separate thread to do the logging, then couldn't that be accomplished with a custom logger that writes the log entries to a pipe and has its own separate thread for doing the actual logging?

sonots:
Basically, mutex#lock can't be called from trap context. ref. https://bugs.ruby-lang.org/issues/7917
Problem is not only for logger.
Pushing signal into an array or pipe in trap context, and hanling them in somewhere else is the common way handling signal as unicorn, sidekiq, fluentd does.

sonots:
* sidekiq https://github.com/mperham/sidekiq/blob/b6de0bae6b86d5abae25852141768d8ecc8ddedf/lib/sidekiq/cli.rb#L53
* unicorn https://github.com/defunkt/unicorn/blob/master/lib/unicorn/http_server.rb#L120

sonots:
We basically have to avoid heavy work in trap context because we can not trap a new signal during processing something inside trap context.
We therefore should only put signals into an array or pipe in trap context, and then treat signals somewhere else sequentially not to lose new signals.

sonots:
I came up with a way not to use thread > https://github.com/resque/resque/compare/master...sonots:fix_signal_trap2?expand=1

However, in this way, the behavior of signal handler for a forked child must have been changed if TERM_CHILD environment variable is not given.

Before:

* TERM: Set `@shutdown = true`, but actually affects nothing
* INT: Set `@shutdown = true`, but actually affects nothing
* QUIT: Set `@shutdown = true`, but actually affects nothing
* USR1: Do nothing
* USR2: run_hook :before_pause and set `@paused = true`
* CONT: Set `@paused = false` and run_hook :after_pause

After:

* All signals are reverted to DEFAULT

Which do you prefer?

BTW: Rather, I did not get why resque was not calling unregister_signal_handlers for a forked child always. I felt as `Before` behavior was wierd because it disabled TERM or INT signals.

fw42:
The problem I see with your other approach is that that code (where the `sleep` was) will only get executed if the queue is empty (`work_one_job` returns `nil`/`false`). So if there is sufficiently many jobs being pushed into the queue, you will never get to the signal handling part.

Can you open a PR for that so that we can discuss the code?

sonots:
Okay, sent another PR https://github.com/resque/resque/pull/1502

sonots:
I added Thread.stop to stop main thread on receiving signal once, not to require handling a signal to be thread-safe anymore > https://github.com/resque/resque/pull/1499/commits/8fff907d49b73602211ef5efbfdd013053848889

lib/resque/worker.rb
```diff
@@ -861,7 +862,11 @@ def perform_with_fork(job, &block)
 
       begin
         @child = fork do
-          unregister_signal_handlers if term_child
+          if term_child
+            unregister_signal_handlers
+          else
+            @signal_handler.reopen
```
fw42: why is this necessary?

sonots: It is because

1. fork does not copy thread
2. IO.pipe is shared with the parent process

which results in that receiving a signal in child process makes an effect to its parent process.
`reopen` newly creates signal handler thread and pipe for the child process.

fw42: Ok makes sense!

fw42: are any tests failing if you remove this line here? If not, can you add a new test?

sonots: It seems difficult to create tests for handling signals, but let me try it later.

sonots: Added https://github.com/resque/resque/pull/1499/commits/885d51fb86651be61bdbb4485bbe7b27df405331, but sleep 0.1 * 10 times would be feasible...

test/child_killing_test.rb
```diff
@@ -92,11 +98,26 @@ def assert_child_not_running(child_pid)
...
+
+    Process.kill('TERM', child_pid)
+    10.times do
+      assert_nil Process.waitpid(worker_pid, Process::WNOHANG) # should not die
+      sleep 0.1
+    end
+
+    Process.kill('TERM', worker_pid)
+    Process.waitpid(worker_pid)
+    result = Resque.redis.lpop('sigterm-test:result')
```
fw42: what does this line do? isn't there an assertion missing here?

test/child_killing_test.rb
```diff
@@ -92,11 +98,26 @@ def assert_child_not_running(child_pid)
...
-      worker_pid, child_pid, result = start_worker(0, true)
+      worker_pid, child_pid, result = start_worker_and_kill_term(0, true)
       assert_equal worker_pid, child_pid, "child_pid should equal worker_pid, since we are not forking"
       assert Resque.redis.lpop( 'sigterm-test:ensure_block_executed' ), 'post cleanup did not occur. SIGKILL sent too early?'
     ensure
       ENV['FORK_PER_JOB'] = old_job_per_fork
     end
   end
+
+  it 'sending signal to a forked child should not make a effect to its parent' do
```
fw42: just to double-check: This test fails without the line above?

sonots: yes

test/child_killing_test.rb
sonots: it seems that LongRunningJob does `rpush`, this line is just for `lpop` it. Is it not needed?

fw42:
I like this approach. @dylanahsmith, @steveklabnik, wdyt?

sonots:
> I added Thread.stop to stop main thread on receiving signal once, not to require handling a signal to be thread-safe anymore > 8fff907

This way was not good. Main thread would already be a stop state before calling 
Thread.stop in some reasons. Main thread stucks if Thread.main.wakeup is called before
Thread.stop. (This situation happened to me). Reverted.

dylanahsmith:
I still dislike this approach.  There seems to be an assumption that we need to move all the signal handling logic outside of the signal handler, but 

> Basically, mutex#lock can't be called from trap context. ref. https://bugs.ruby-lang.org/issues/7917
> Problem is not only for logger.

Where else do we use mutex#lock other than in the implementation of the logger?  The signal handlers are not actually that complicated.  Some of those mostly just set an instance variable, which seems like overkill to do by piping to another thread.

The other possible concern might be the before_pause and after_pause hooks, but moving that to a separate thread could break applications using that hook.  If the application wants to avoid this problem, then they could also avoid logging or just delay logging.

What if we just append those messages to an array, which is safe to do in a signal handler, then pop them off and write them in the main thread.  For instance we could do that when methods like Resque::Worker#shutdown? or Resque::Worker#paused? are called to check those instance variables so that we log before the signal actually has an effect.

chrisccerami:
Is this PR still in progress?

dchersey:
Sorry, did this PR get merged?

PR #1501, opened 2016-08-17
fix tests

fw42:
tests are broken again because of gem updates that don't support old Ruby versions anymore

@dylanahsmith quick look please?


dylanahsmith:
:+1:

PR #1502, opened 2016-08-22
Fix log writing failed. can't be called from trap context (no thread version)

sonots:
Fix https://github.com/resque/resque/issues/1493, similarly with https://github.com/resque/resque/pull/1499, but this is no thread version.

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -228,7 +228,7 @@ module Resque
           break if interval.zero?
           log_with_severity :debug, "Sleeping for #{interval} seconds"
           procline paused? ? "Paused" : "Waiting for #{queues.join(',')}"
-          sleep interval
+          @signal_handler.handle_signal(interval)
         end
       end
 
@@ -394,11 +394,15 @@ module Resque
     end
 
     def unregister_signal_handlers
-      trap('TERM') do
-        trap ('TERM') do
-          # ignore subsequent terms
+      if term_child
+        trap('TERM') do
+          trap ('TERM') do
+            # ignore subsequent terms
+          end
+          raise TermException.new("SIGTERM")
         end
-        raise TermException.new("SIGTERM")
+      else
+        trap('TERM', 'DEFAULT')
       end
       trap('INT', 'DEFAULT')
 
@@ -862,11 +866,7 @@ module Resque
 
       begin
         @child = fork do
-          if term_child
-            unregister_signal_handlers
-          else
-            @signal_handler.reopen
-          end
+          unregister_signal_handlers
           perform(job, &block)
           exit! unless run_at_exit_hooks
         end
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -373,18 +373,19 @@ module Resque
     # USR2: Don't process any new jobs
     # CONT: Start processing jobs again after a USR2
     def register_signal_handlers
-      trap('TERM') { graceful_term ? shutdown : shutdown!  }
-      trap('INT')  { shutdown!  }
+      st = Resque::SignalTrap.new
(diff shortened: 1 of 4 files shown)


lib/resque/worker.rb
```diff
@@ -228,7 +228,7 @@ def work(interval = 5.0, &block)
           break if interval.zero?
           log_with_severity :debug, "Sleeping for #{interval} seconds"
           procline paused? ? "Paused" : "Waiting for #{queues.join(',')}"
-          sleep interval
+          @signal_handler.handle_signal(interval)
```
fw42: As mentioned in the other PR, this will only run if the queue was empty. Should we move this out of the `unless`?

lib/resque/worker.rb
```diff
@@ -861,7 +866,7 @@ def perform_with_fork(job, &block)
 
       begin
         @child = fork do
-          unregister_signal_handlers if term_child
+          unregister_signal_handlers
```
fw42: can you explain this change?

sonots: This is as I explained in #1499. Below is what I wrote.

---------------------

However, in this way, the behavior of signal handler for a forked child must have been changed if TERM_CHILD environment variable is not given because there is no chance to call `handle_signal(timeout)` in the forked child.

Before:

* TERM: Set `@shutdown = true`, but actually affects nothing
* INT: Set `@shutdown = true`, but actually affects nothing
* QUIT: Set `@shutdown = true`, but actually affects nothing
* USR1: Do nothing
* USR2: run_hook :before_pause and set `@paused = true`
* CONT: Set `@paused = false` and run_hook :after_pause

After:

* All signals are reverted to DEFAULT

Which do you prefer the thread version or no thread version?

BTW: Rather, I did not get why resque was not calling unregister_signal_handlers for a forked child always. I felt as `Before` behavior was wierd because it disabled TERM or INT signals.

lib/resque/worker.rb
sonots: Hmm, if we move this out of the unless, waiting interval is issued even if we run a job (behavior changes).
I am not getting good idea yet. I will think of how to resolve this. Any idea is welcome.

fw42: can we do a non-blocking `select` so it doesn't sleep and just returns if there was no signal yet? and keep the `sleep` here where it was

sonots: applied a change like https://github.com/resque/resque/pull/1502/commits/4ac5b36b7eb3f6cb0a58c40f41af47e6ea25a4a4.

> can we do a non-blocking select so it doesn't sleep and just returns if there was no signal yet? and keep the sleep here where it was

During sleeping, we can not get signal, then. I felt `@signal_handler.handle_signal(interval)` here is still better, but I added `@signal_handler.handle_signal(0)` for the case `work_one_job` returns true.

Also, I added `handle_signal` in `perform_with_fork` for during waiting a forked child finishes.

Please note that we still have problems that we can not handle signal `fork_per_job?` is false. Hmm.

lib/resque/worker.rb
```diff
@@ -224,11 +224,13 @@ def work(interval = 5.0, &block)
       loop do
         break if shutdown?
 
-        unless work_one_job(&block)
+        if work_one_job(&block)
+          @signal_handler.handle_signal(0)
+        else
           break if interval.zero?
```
fw42: this line here might mess up your logic.. if the interval is zero and the queue is empty, then we wont handle signals...

sonots: thank you for pointing out

sonots:
We still have problems that we can not handle signal during processing a job when fork_per_job? returns false. Thread vesion does not have such an issue.

sonots:
As a final thought, there is no way to resolve the above issue with this no thread version. Let me close.

PR #1506, opened 2016-09-01
Change $?.stopsig to $?.inspect on DirtyExit

sonots:
$?.stopsig returns the number of the signal that caused stat to stop (or
nil if self is not stopped). The **stop** does not include **terminate**.
So, we get nil at this line usually.
In the case of **terminate**, termsig is available.

$?.inpect is most elaborative for both case

ref. http://docs.ruby-lang.org/en/2.0.0/Process/Status.html#method-i-stopsig

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -880,7 +880,7 @@ module Resque
         nil
       end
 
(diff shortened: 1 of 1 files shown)


sonots:
Ah, I missed https://github.com/resque/resque/pull/1503. #1503 is fine for me too.

PR #1511, opened 2016-09-22
Use notify_sync when calling Airbrake

earnold:
Inside of Airbrake, for the class Resque::Failure::Airbrake, they use the method `notify_sync` (https://github.com/airbrake/airbrake/blob/master/lib/airbrake/resque/failure.rb), while inside of Resque, it calls `notify`.  I believe that the Airbrake behavior is correct.

Here's my thinking:  under the hood of Airbrake, `notify` uses a thread pool of async
notifiers. This appears to be incompatible with Resque and errors
are being swallowed.

### DIFF
--- a/lib/resque/failure/airbrake.rb
+++ b/lib/resque/failure/airbrake.rb
@@ -21,7 +21,7 @@ module Resque
       end
 
       def save
-        ::Airbrake.notify(exception,
+        ::Airbrake.notify_sync(exception,
             :parameters => {
             :payload_class => payload['class'].to_s,
             :payload_args => payload['args'].inspect


chrisccerami:
Can you describe in a bit more detail the problem you believe this PR solves?

earnold:
@chrisccerami The problem, writ large, is that that calling Airbrake from Resque jobs does not work when you call `notify`.  I've had to monkey patch this in a project of mine to get it working. 

Fairly easy to reproduce: you just need a Resque job that throws an error.

chrisccerami:
It seems that the version of Airbrake used by Ruby 1.8.7 doesn't support this. I have opened an issue to discuss dropping support for older Ruby versions, but it hasn't received much conversation as of yet, so I don't know where that stands.

chrisccerami:
The removal of 1.8.7 and 1.9.3 support has now been merged into master, so if you rebase this PR it should be good to go.

Tonkpils:
@earnold any update on the rebase?

jeremy:
Superseded by #1581

PR #1205, opened 2014-04-15
Implemented pre_term_timeout including test

ejlangev:
Implements the same functionality but includes a test of the signal handling to make sure things are working properly.

See https://github.com/resque/resque/pull/1136 for a description of the use case for this code.

PR #1514, opened 2016-09-26
Add pre_term_timeout

tskogberg:
This is all of https://github.com/resque/resque/pull/1205 squished into
one and then rebased on the latest resque master.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -449,6 +449,22 @@ If you want to stop processing jobs, but want to leave the worker running
 (for example, to temporarily alleviate load), use `USR2` to stop processing,
 then `CONT` to start it again.
 
+#### Signals on Heroku
+
+When shutting down processes, Heroku sends every process a TERM signal at the
+same time. By default this causes an immediate shutdown of any running job
+leading to frequent `Resque::TermException` errors.  For short running jobs, a simple
+solution is to give a small amount of time for the job to finish
+before killing it.
+
+To accomplish this set the following environment variables:
+
+* `RESQUE_PRE_SHUTDOWN_TIMEOUT` - The time between the parent receiving a shutdown signal (TERM by default) and it sending that signal on to the child process. Designed to give the child process
+time to complete before being forced to die.
+
+* `RESQUE_SHUTDOWN_SIGNAL` - The signal sent to the child process
+when the parent receives any exit signal.
+
 ### Mysql::Error: MySQL server has gone away
 
 If your workers remain idle for too long they may lose their MySQL connection. Depending on your version of Rails, we recommend the following:
--- a/README.markdown
+++ b/README.markdown
@@ -462,8 +462,9 @@ To accomplish this set the following environment variables:
 * `RESQUE_PRE_SHUTDOWN_TIMEOUT` - The time between the parent receiving a shutdown signal (TERM by default) and it sending that signal on to the child process. Designed to give the child process
 time to complete before being forced to die.
 
-* `RESQUE_SHUTDOWN_SIGNAL` - The signal sent to the child process
-when the parent receives any exit signal.
+* `TERM_CHILD` - Must be set for `RESQUE_PRE_SHUTDOWN_TIMEOUT` to be used. After the timeout, if the child is still running it will raise a `Resque::TermException` and exit.
+
+* `RESQUE_TERM_TIMEOUT` - By default you have a few seconds to handle `Resque::TermException` in your job. `RESQUE_TERM_TIMEOUT` and `RESQUE_PRE_SHUTDOWN_TIMEOUT` must be lower than the [heroku dyno timeout](https://devcenter.heroku.com/articles/limits#exit-timeout).
 
 ### Mysql::Error: MySQL server has gone away
 
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -400,15 +400,15 @@ module Resque
     def unregister_signal_handlers
       trap('TERM') do
         trap('TERM') do
-          # Ignore subsequent shutdown signals
+          # Ignore subsequent term signals
         end
 
-        log_with_severity :debug, "Trapped TERM in child #{Process.pid}; raising"
         raise TermException.new("SIGTERM")
       end
 
+      trap('INT', 'DEFAULT')
+
       begin
-        trap('INT', 'DEFAULT')
         trap('QUIT', 'DEFAULT')
         trap('USR1', 'DEFAULT')
         trap('USR2', 'DEFAULT')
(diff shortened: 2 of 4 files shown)


tskogberg:
Travis CI seems a bit unstable. This code/tests run fine locally. If you have more access to Travis please re-run the tests again. 👍

README.markdown
```diff
@@ -449,6 +449,22 @@ If you want to stop processing jobs, but want to leave the worker running
...
+leading to frequent `Resque::TermException` errors.  For short running jobs, a simple
+solution is to give a small amount of time for the job to finish
+before killing it.
+
+To accomplish this set the following environment variables:
+
+* `RESQUE_PRE_SHUTDOWN_TIMEOUT` - The time between the parent receiving a shutdown signal (TERM by default) and it sending that signal on to the child process. Designed to give the child process
+time to complete before being forced to die.
+
+* `RESQUE_SHUTDOWN_SIGNAL` - The signal sent to the child process
```
fw42: why do we need the possibility to overwrite the shutdown signal? Does Heroku not use SIGTERM?

test/worker_test.rb
```diff
@@ -1139,6 +1137,80 @@ def run_in_job(&block)
...
+          begin
+            pre_shutdown_timeout = 1
+            Resque.enqueue(PreShutdownLongRunningJob, run_time)
+
+            worker_pid = Kernel.fork do
+              # reconnect to redis
+              Resque.redis.client.reconnect
+
+              # ensure we fork (in worker)
+              $TESTING = false
```
fw42: looks like this variable will leak into other tests... seems a bit sketchy

joakimk:
In case anyone wants to use this on top of the latest stable version we've got a branch for that:

`gem "resque", ">= 1.26.0", require: "resque/server", github: "barsoom/resque", branch: "v1.26.0_with_pre_term_timeout"`

README.markdown
joakimk: I don't really know, this is basically just the code @ejlangev wrote. Do you have any input Ethan?

test/worker_test.rb
joakimk: Does not seem to be used for anything, removed it.

README.markdown
ejlangev: I honestly don't remember that well.  I wrote this code 2 years ago at this point.  I think it was to make the behavior backwards compatible.  Here's the blog post I wrote explaining the issue: http://eng.joingrouper.com/blog/2014/06/27/too-many-signals-resque-on-heroku/.  Based on re-reading that I think the child process needs to be able to ignore SIGTERM because heroku sends it that signal earlier than it should.

joakimk: We do not seem to need it, this is the setup we have now. 

`resque:        env RESQUE_PRE_SHUTDOWN_TIMEOUT=25 RESQUE_TERM_TIMEOUT=2 TERM_CHILD=1 INTERVAL=1 QUEUE=high,medium,mail,low bundle exec rake resque:work`

It allows us 25 seconds for jobs to finish, then 2 seconds (as far as we understand the code) for the failure handling to finish. We use that time to requeue all jobs that haven't finished by catching TermException in `Job#fail`(so that we don't have to catch it in all of our jobs).

We haven't actually run it on heroku yet though, but our old environment is setup to send the same signals. TERM first, then wait 30 seconds and KILL if it's still running. We're in the process of migrating to heroku.

fw42: if it's not needed, maybe we should remove it?

ejlangev: Make sure in testing you send the TERM signal to both the parent worker and the child process that actually executes the job.  That was the crux of the issue.  Just sending it to the parent process already has working behavior without this change but doesn't match what actually happens on heroku.

tskogberg: We removed the config and ensured it works as it should on heroku. In order for it to work `TERM_CHILD` needs to be set. We updated the readme to reflect that.

When heroku sends a TERM to the parent process, it waits for `RESQUE_PRE_SHUTDOWN_TIMEOUT`, then sends TERM to the child, and waits `RESQUE_TERM_TIMEOUT` for it to handle `Resque::TermException`. This works as long as both timeouts combines is less than the heroku timeout before it sends KILL (currently 30 seconds).

ejlangev: Okay, it's possible heroku changed somewhat since this code was written.  I recall at the time that it was sending TERM To parent and child at the same time and that was the crux of the problem here.  The parent could wait for some time before sending a term to the child but it didn't matter because heroku sent the signal to both.

joakimk: We've been handing `Resque::DirtyExit` by requeuing jobs just as we do with `Resque::TermException`, but this could be caused by not handling this signal correctly. We'll try and see tomorrow if we fix it propertly so we don't need to catch DirtyExit.

joakimk: I think I've found the issue we had and it's a server configuration issue, not a bug in this code. I'll report back when we've tried it out properly tomorrow.

tskogberg: I can confirm that the issue we hade with `Resque::DirtyExit` was no bug in resque.

Our current servers didn't behave like heroku would, during a restart it was possible to kill new workers that was booting up. So had nothing to do with resque! Besides that self inflicted issue we have found no others regarding this pull requests unless you have found anything else.

joakimk:
Now the build works. So as a bonus for merging this, master will also have a working build :)

Btw, is there any settings for that coverage bot? Lowest coverage change to notify about or something like that? It's kinda spammy.

chrisccerami:
My only nitpick is that the comment on lines 516-518 of worker.rb should be updated to more accurately describe the added behavior to `new_kill_child`. Otherwise this seems reasonable to me.

chrisccerami:
Seeing as how I don't notice any breaking changes here, I'm going to merge this now and I'll open a PR to fix the comment I mentioned above.

robinmonjo:
Thank you for this feature !! Any idea when a new release including it will be available ?

lib/resque/worker.rb
```diff
@@ -394,11 +399,13 @@ def register_signal_handlers
 
     def unregister_signal_handlers
       trap('TERM') do
-        trap ('TERM') do
-          # ignore subsequent terms
+        trap('TERM') do
+          # Ignore subsequent term signals
         end
+
         raise TermException.new("SIGTERM")
```
iloveitaly: @chrisccerami This `TermException` seems to be causing issues in my case. If `fork_per_job? && term_child` then this code is run on the child process which causes the child process to immediately exit with a `TermException` without waiting for the second `SIGTERM` initiated by the parent after `RESQUE_PRE_SHUTDOWN_TIMEOUT`

PR #1515, opened 2016-09-27
Fix warnings and deprications

tskogberg:
Generally this will make running the tests and using resque easier since
the console isn't flooded with warnings.
- Overshadowing variables in templates
  - Warning because you might use the wrong variable.
- Redefinition of constants
  - Warning because it might not be what you intended.
- Unused variables (e.g. _foo)
  - Warning because you might expect them to be used.
- Usage of undefined instance variables
  - Warning because doing so could be because of a typo.

There are still warnings but they are harder to fix. For example, the
warning about setting a global value on Encoding should probably not
be hidden and we don't know how to work around it right now.

The monkeypatch override of UTF handling in 1.9 can't really be fixed
without a redesign.

And probably a few other things as well.


tskogberg:
We ran the tests using `REDIS_NAMESPACE_QUIET=true rake` since we didn't try to fix that issue. Without that it's hard to see the improvements.

lib/resque.rb
```diff
@@ -550,26 +550,23 @@ def sample_queues(sample_size = 1000)
...
+  # Retrieve all hooks of a given name.
+  def all_hooks_by_name(name)
+    hooks[name] || []
   end
 
-  # Retrieve all hooks of a given name.
-  def hooks(name)
-    (@hooks && @hooks[name]) || []
+  # Retrieve all hooks
+  def hooks
```
fw42: Doesn't it break backwards compatibility to change this method from `hooks(name)` to `hooks()`?

test/child_killing_test.rb
```diff
@@ -62,13 +62,13 @@ def assert_child_not_running(child_pid)
 
   if !defined?(RUBY_ENGINE) || RUBY_ENGINE != "jruby"
     it "old signal handling just kills off the child" do
-      worker_pid, child_pid, result = start_worker(0, false)
+      _worker_pid, child_pid, result = start_worker(0, false)
```
fw42: what is it that generates warnings about this? Ruby itself doesn't, does it? Is this a Rubocop thing?

lib/resque.rb
joakimk: It's a private method. We assumed that breaking that kind of backward compatibility would be fine since no one should be using it?

test/child_killing_test.rb
joakimk: At least in ruby 2.3.0 when run with the "-w" flag, unused variables with get a warning like: "lab.rb:2: warning: assigned but unused variable - bar".

The test suite seems to run with warnings enabled. Which seems sensible.

teeparham:
There are too many changes in one commit here. It should be broken into smaller commits.

chrisccerami:
I think it's fine as one larger commit, I don't see an issue with it. This also seems like a sensible change to me. 👍

fw42:
@hkdsun, mind taking a quick look?

lib/resque/server.rb
```diff
@@ -131,19 +131,23 @@ def partial(template, local_vars = {})
...
 
+      def polling?
+        @polling ||= false
+      end
     end
 
     def show(page, layout = true)
       response["Cache-Control"] = "max-age=0, private, must-revalidate"
       begin
+        @subtabs = nil
```
hkdsun: Why do we need this?

lib/resque/worker.rb
```diff
@@ -619,6 +621,8 @@ def run_hook(name, *args)
     end
 
     def kill_background_threads
+      @heartbeat_thread ||= nil
```
hkdsun: I like the pattern of having a reader method on these uninitialized instance variables (doing it in constructor is okay too). It makes it clearer what the default initialization value is. Is there a reason why we shouldn't do that for this variable?

lib/resque/failure.rb
```diff
@@ -31,6 +31,7 @@ def self.backend=(backend)
     # Returns the current backend class. If none has been set, falls
     # back to `Resque::Failure::Redis`
     def self.backend
+      @backend ||= nil
```
hkdsun: Same comment as the `@heartbeat_thread` variable. Is it possible that we move this initialization to somewhere more obvious?

lib/resque.rb
```diff
@@ -483,7 +483,7 @@ def remove_worker(worker_id)
   # Returns a hash, similar to redis-rb's #info, of interesting stats.
   def info
     return {
-      :pending   => queue_sizes.inject(0) { |sum, (queue_name, queue_size)| sum + queue_size },
+      :pending   => queue_sizes.inject(0) { |sum, (_queue_name, queue_size)| sum + queue_size },
```
chrisccerami: Isn't `_queue_name` an unused variable? If it's unused, usually the convention is to name it `_`, no?

chrisccerami:
Just one small comment about an unused variable.

PR #1516, opened 2016-09-28
Fix bug when encontering an error while pruning workers

tskogberg:
The error handling while pruning workers had a bug.

In our case this was our temporary monkeypatch of `Job#fail` caused an error.

Regardless of that, this error reporting code was broken and should be fixed.

/ tomas and joakim

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem "i18n"
 if RUBY_VERSION < "2.0"
   gem "json", '~> 1.8'
   gem "coveralls", "0.8.13", :require => false
+  gem "term-ansicolor", "1.3.2" # A dependency of coveralls. The next version requires ruby >= 2.0.
 else
   gem "json"
   gem "coveralls", :require => false
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -650,8 +650,8 @@ module Resque
     rescue Exception => exception_while_unregistering
       message = exception_while_unregistering.message
       if exception
-        message = message + "\nOriginal Exception (#{exception.class}): #{exception.message}\n" +
-                            "  #{(exception.backtrace || []).join("  \n")}"
+        message += "\nOriginal Exception (#{exception.class}): #{exception.message}"
+        message += "\n  #{exception.backtrace.join("  \n")}" if exception.backtrace
       end
       fail(exception_while_unregistering.class,
            message,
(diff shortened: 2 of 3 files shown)


lib/resque/worker.rb
```diff
@@ -651,7 +651,7 @@ def unregister_worker(exception = nil)
       message = exception_while_unregistering.message
       if exception
         message = message + "\nOriginal Exception (#{exception.class}): #{exception.message}\n" +
-                            "  #{exception.backtrace.join("  \n")}"
+                            "  #{(exception.backtrace || []).join("  \n")}"
```
fw42: Hmm interesting. Not sure I understand why this is necessary. Some exceptions don't have backtraces?

Gemfile
```diff
@@ -8,6 +8,7 @@ gem "i18n"
 if RUBY_VERSION < "2.0"
   gem "json", '~> 1.8'
   gem "coveralls", "0.8.13", :require => false
+  gem "term-ansicolor", "1.3.2" # A dependency of coveralls. The next version requires ruby >= 2.0.
```
fw42: Was this the reason why the tests were failing? Thanks for fixing it!

test/worker_test.rb
```diff
@@ -744,6 +744,24 @@ def self.perform
     end
   end
 
+  it "reports errors correctly when an error occurs while pruning workers" do
```
fw42: maybe the name of the test could be more explicit, like "correctly reports errors which don't have a backtrace"?

lib/resque/worker.rb
joakimk: The `PruneDeadWorkerDirtyExit` exception does not since it's never thown, only created at https://github.com/resque/resque/blob/master/lib/resque/worker.rb#L578

Gemfile
joakimk: It seems to be the reason, but this didn't fix it. We gave up at this point since we don't have direct access to your CI server.

test/worker_test.rb
joakimk: I pushed a fix.

fw42:
Cool. Thanks for the fix! This looks reasonable to me. @dylanahsmith, mind taking a look?

lib/resque/worker.rb
dylanahsmith: There is no need for "\n  " on the end in this case.  I would instead this and the above line to

```ruby
message += "\nOriginal Exception (#{exception.class}): #{exception.message}"
message += "\n  #{exception.backtrace.join("  \n")}" if exception.backtrace
```

dylanahsmith:
Other than the minor nitpick, LGTM

lib/resque/worker.rb
joakimk: Makes sense, changed it to that instead.

fw42:
Thanks for your contribution!

Issue #1526, opened 2016-11-27
What Ruby versions should we continue supporting?


chrisccerami:
I figured I'd open an issue for this since it's probably the easiest way to track discussion on it. I've been trying to decide if I think it makes any sense for Resque to continue actively supporting some of the older EOL Rubies. I feel like 1.8.7 is so far past its prime that it's probably putting more of a burden on contributors and maintainers than any potential benefit of pushing upgrades to 1.8.7 users, so in my opinion, dropping 1.8.7 seems pretty reasonable.

1.9.3 is similarly EOL, but I'm not sure how large of a base there is of people still using it, and a similar point can be made about 2.0.x, which was EOL'd in February of this year.

So the main point of contention, I suppose, is do we keep updating Resque for EOL Rubies, or do we just allow those users to lock their version at `1.26.0` and move on to focusing on supported versions?

rjhancock:
I'd suggest dropping support for any version of Ruby that is no longer being supported.

steveklabnik:
Historically speaking, a lot of resque's value is in legacy. That doesn't mean that every Ruby till the dawn of time should be supported, but it'd be nice to try and figure out some way to figure out what Rubies resque users use.

I'd also recommend that moving to a 2.x is a good idea if 1.8 support is gonna be dropped.

rjhancock:
Then I'd stop at 2.0. Granted I'm working on a project that still runs on 1.8.7, but it's slowly moving to 2.0 (then newer after that).

chrisccerami:
Yeah, I'm certainly not trying to argue that Resque should only support latest and greatest. I agree Steve, that it would be great if we had some kind of stats on what Rubies Resque users are on.

IAMALLWE:
I Know WHY family, I Forgive IGNORANCE. PRIDE is a mocker, strong drink a brawler, and whoEvEr is led astray by them Is A FOOL too. Bf4ore Wise there First must be Fool (Jester) Be ALL things to ALL people. You SEE NAMASTE PEACE

pboling:
I have a large resque installation on ruby-2.0.0, and will likely be for another few months.  I think stopping at 2.0.0 is reasonably (biased!).

wjessop:
+1 for Ruby 2 minimum. Anything older is ancient.

IAMALLWE:
🎹

claui:
I’d love for Resque to continue supporting **Ruby 2.0.0 and above.**

[...]

PR #1535, opened 2017-02-13
Remove support for Rubies < 2.0.0

chrisccerami:
With this PR, I'm proposing that we officially remove support for any Rubies older than 2.1.10. My reasoning for this is that while legacy support has and will be an important goal for Resque, we simply can't support every Ruby throughout history.

Ruby 1.8.7 was released in 2008 and End-of-Lifed in 2013.

Ruby 1.9.3 was released in 2011 and EOLed in 2014.

Even 2.0.0 was released in 2013 and was EOLed at the end of 2015.

While Ruby 2.1.10 will be EOLed soon, I think that for now it is the most reasonable minimum Ruby version for any further releases of Resque. I hope to hear from the rest of the Resque team and those using Resque to gain some sort of consensus on this point.


chrisccerami:
Related to #1526

steveklabnik:
Not that I have a ton of weight in this, but I feel like dropping support for these rubies is worth a 2.0 release. Something to consider; please don't consider this opinion blocking.

chrisccerami:
@steveklabnik I don't disagree with you. I think it could be seen as a breaking change.

oss92:
You will need to remove 
```
if RUBY_VERSION <= '1.9'
  require 'resque/vendor/utf8_util/utf8_util_18'
else
  require 'resque/vendor/utf8_util/utf8_util_19'
end
```

and remove `resque/vendor/utf8_util/utf8_util_18` and `resque/vendor/utf8_util/utf8_util_19` after copying the `clean!` method from the latter.

pboling:
Dropping support for any Ruby version is one of the best reasons for a major version bump.  See SemVer 2.0#8 -
> 8. Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. It MAY include minor and patch level changes. Patch and minor version MUST be reset to 0 when major version is incremented.

Source: http://semver.org/

Regardless of whether or not this project follows all of the principles of SemVer or not, there are good reasons for this rule.

chrisccerami:
Sure, I think we all agree with that.

chrisccerami:
I think we've collected enough feedback to say that people are probably comfortable with removing support for < 2.0.0, so I'm going to merge this into master. I'm not going to release it right away, as this will constitute a 2.x version of Resque, and we may want to make other changes at the same time.

PR #1536, opened 2017-02-13
CHANGE: Adjust header namespace color to be obvious

oss92:
After:
screen shot 2017-02-13 at 19 06 35
Before:
screen shot 2017-02-13 at 19 09 00


chrisccerami:
That does seem like an improvement. Thanks!

Issue #1537, opened 2017-02-14
Redis 2.4: Redis::CommandError: ERR unknown command 'time'


tf:
[This recent change](https://github.com/resque/resque/commit/47bce253ebc23df93329de4979392a065b701b3b#diff-102541f6f2ba476fceb8e45917e63420R93) by @fw42 appears to be breaking Resque on Redis 2.4 for me.  Though I do not know the details of the change, the `time` command is [only available as of Redis 2.6](https://redis.io/commands/time). Has support for Redis 2.4 been dropped? Where do find out more about Resque's supported Redis versions?

chrisccerami:
Hm, I don't think support was intentionally dropped but Redis 2.4 is a pretty old Redis. Redis 2.6 was released in 2012, and Redis generally doesn't support for more than one stable release back. Current stable Redis is 3.2, which means that back to 3.0 is supported by the Redis team.

We should probably keep track of what versions of Redis will work with Resque, but I don't think anyone has done that work.

tf:
Engine Yard provides Redis 2.4 by default as part of their v4 stack, so I
guess it is still in use. We will probably have to use an old resque
version then.

Chris C Cerami <notifications@github.com> schrieb am Di., 14. Feb. 2017,
21:35:

chrisccerami:
If Redis 2.4 is still widely used, we can also attempt to fix that problem in a future patch version of Resque. I don't know what exactly a potential solution would be, but I'd think @fw42 would know if there is a fix.

danbernier:
I'm currently seeing this with Redis 3.2.6, using the redis gem 3.3.3.

chrisccerami:
That's interesting. There is definitely a TIME command as of Redis 2.6.0, as seen [here](https://redis.io/commands/time), so if this issue is happening on newer Redis then something is clearly wrong. I'll try to take a look at it today.

danbernier:
Ah - I stand corrected, apologies. The server I'm seeing this issue on is, in fact, running 2.4. Urgency downgraded.

chrisccerami:
That's a relief 😆 this almost became a much weirder issue.

chrisccerami:
[...]

PR #1538, opened 2017-02-15
Fallback to server time if Redis time won't work

chrisccerami:
This addresses #1537. If `@redis.time` raises a `Redis::CommandError`, then the `DataStore` will fall back to `Time.now`.


oss92:
Two points
1. Why is it necessary to call `@redis.time` twice? Once for checking and then once for returning.
2. There is a deprecation warning in the `server_time` method as well. If the Redis connection is namespaced then the following occurs
```
Passing 'time' command to redis as is; blind passthrough has been deprecated and
will be removed in redis-namespace 2.0
(at /.../gems/resque-1.27.1/lib/resque/data_store.rb:94:in `server_time')
```

which can be resolved by checking whether the Redis connection is namespaced so `if @redis.responds_to?(:redis)` we should retrieve the time by `@redis.redis.time`.

chrisccerami:
For point 1, it's called first on initialization, to check if we can use Redis time or if we have to use server time, and then it's called many times after by the heartbeat code. The reason I'm checking on initialize is so that we don't need to rescue an exception every single time we want to check the time, which would end up being thousands and thousands of exceptions to rescue. It's much more efficient to check if we can once, and save that value.

For point 2, that's a good call. I'll make that update.

chrisccerami:
`redis-namespace` 1.5.3 has now been released and should fix the blind passthrough of `time`, so that shouldn't be an issue anymore.

fw42:
As far as I remember, using the server time (instead of the client time) fixed some kind of race condition. Will this fallback reintroduce that bug?

chrisccerami:
@fw42 it's likely that it will introduce that bug, but currently Resque is broken for Redis 2.4 users, so it's kind of a tradeoff. If you're using a modern Redis (>2.4) you'll get `Redis.time`, otherwise yes, the bug will probably show back up. We could alternately decide to not support Redis 2.4, but that seems like an abrupt decision to make. I'd appreciate your input on the subject.

lib/resque/data_store.rb
```diff
@@ -91,10 +92,16 @@ def all_resque_keys
     end
 
     def server_time
-      time, _ = @redis.time
+      time, _ = @redis_time_available ? @redis.time : Time.now
       Time.at(time)
     end
 
+    def redis_time_available?
+      @redis.time
```
fw42: does `@redis.respond_to?(:time)` work too?

lib/resque/data_store.rb
```diff
@@ -91,10 +92,16 @@ def all_resque_keys
...
+    rescue Redis::CommandError
```
fw42: Does this error come from the server? i.e. do we need to send this command to the server in order to know that it's not supported? should we cache the response so that we don't have to ping the server every time?

lib/resque/data_store.rb
chrisccerami: I think `@redis.respond_to?(:time)` will only work if Redis isn't namespaced. Otherwise we would need to check `@redis.redis.respond_to?(:time)`, and I figured it would be more effective to just try to call `time` and let it fail if it doesn't exist.

lib/resque/data_store.rb
chrisccerami: The response is cached when the DataStore is initialized, so as long as you're using the same DataStore model it won't call this method again. I should probably also make this method private though.

Issue #1544, opened 2017-02-23
Resque 1.27.2 broke Resque.inline.


hodgestar:
It looks like a change between 1.27.1 and 1.27.2 caused `Resque.redis = "redis://..."` to start connecting to redis even when `Resque.inline = true`.

Was this expected? If so, is there a good way to avoiding connecting to redis when in inline mode?

We're using inline mode in our test suite.

chrisccerami:
That definitely doesn't sound intentional. I'll try to look into it today.

chrisccerami:
I think I know what the issue is here, but I haven't had the time to test it yet. I think the problem is caused by `DataStore` now trying to call `redis.time` at initialize time in order to check if it has that command available. I'm not 100% sure yet, but I'd bet this is where it's mistakenly trying to connect to Redis even when inline. Do you have any more information about this issue, @hodgestar? Maybe the output of a test failure or an error with a stacktrace?

hodgestar:
Partial traceback during test run with `Resque.inline = true`:
```
expected ReceiverError with "redis is pining for the fjords", got #<SocketError: getaddrinfo: Name or service not known> with backtrace:
         # /home/simon/.rvm/gems/ruby-2.3.1@pdg-data-receiver/gems/redis-3.3.3/lib/redis/connection/ruby.rb:229:in `getaddrinfo'
         # /home/simon/.rvm/gems/ruby-2.3.1@pdg-data-receiver/gems/redis-3.3.3/lib/redis/connection/ruby.rb:229:in `connect'
         # /home/simon/.rvm/gems/ruby-2.3.1@pdg-data-receiver/gems/redis-3.3.3/lib/redis/connection/ruby.rb:313:in `connect'
         # /home/simon/.rvm/gems/ruby-2.3.1@pdg-data-receiver/gems/redis-3.3.3/lib/redis/client.rb:336:in `establish_connection'
         # /home/simon/.rvm/gems/ruby-2.3.1@pdg-data-receiver/gems/redis-3.3.3/lib/redis/client.rb:101:in `block in connect'
         # /home/simon/.rvm/gems/ruby-2.3.1@pdg-data-receiver/gems/redis-3.3.3/lib/redis/client.rb:293:in `with_reconnect'
         # /home/simon/.rvm/gems/ruby-2.3.1@pdg-data-receiver/gems/redis-3.3.3/lib/redis/client.rb:100:in `connect'
[...]

PR #1545, opened 2017-02-24
Resque.inline should prevent hitting Redis

chrisccerami:
This addresses #1544

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -95,6 +95,7 @@ module Resque
       Time.at(time)
     end
 
+    # only needed until support for Redis 2.4 is removed
     def redis_time_available?
       if @redis_time_available.nil? && !Resque.inline
         begin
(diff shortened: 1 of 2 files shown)


lib/resque/data_store.rb
```diff
@@ -97,7 +97,7 @@ def server_time
     end
 
     def redis_time_available?
-      @redis.time
+      @redis.time unless Resque.inline
```
hodgestar: This returns null when inline is true, yes? Is it okay to return null instead of false? Otherwise looks good to me.

chrisccerami: Yup, because `nil` is a falsey value in Ruby.

fw42: Can you explain the consequences of this change? Does this mean that the server-time will never be used (even if the server supports it) unless you have inlining disabled?

chrisccerami: Yes. From my understanding (and I may be wrong) when `Resque.inline` is set to true, Redis is bypassed entirely and all workers are run in the foreground.

chrisccerami:
Should also address #1552

PR #1546, opened 2017-02-27
Add ActiveJob section to the README

etdsoft:
I was going to ask this in the mailing list but librelist seems down so I thought, why not PR-it?

We used to have all our jobs inherit from a BaseWorker that would call `ActiveRecord::Base.clear_active_connections!` before calling the child's job `#perform` method. Would the approach suggedted in this PR be the recommended approach when using the `:resque` adapter for ActiveJob provided all jobs need the DB, and you're running MySQL?

Thanks!


rafaelfranca:
Can you explain why is this needed for all the jobs? AFAIK you should not have to do this.

PR #1550, opened 2017-03-08
support calling select on redis data_store

allenlsy:


lib/resque/data_store.rb
```diff
@@ -57,7 +57,11 @@ def initialize(redis)
     # Compatibility with any non-Resque classes that were using Resque.redis as a way to access Redis
     def method_missing(sym,*args,&block)
       # TODO: deprecation warning?
-      @redis.send(sym,*args,&block)
+      if sym == :select
+        @redis.select(args.first)
```
chrisccerami: I don't understand why you can't send the `select` method that same way as all others.

allenlsy:
Can you try it? It raises an error on my machine

allenlsy:
@chrisccerami I'm still looking for the reason why it's suddenly happen, but I think 1) it's not the problem of resque 2) we should not monkey-patch it. So I'm closing the PR now.

In my project, I'm using resque/redis-namespace. The redis I'm using has a namespace. I'm not sure whether it's related. just FYI

PR #1551, opened 2017-03-09
Fixed filter log statement to allow for resque job.rb patching. 

tmedford:
Removed lib dependency. Causes error when  gem/lib/resque/job.rb where paths like ./lib/extensions/resque/job.rb are in play


tmedford:
Any update

tmedford:
Any update

chrisccerami:
This seems reasonable enough to me, although I don't entirely understand the issue it's solving. Could you just please add an entry to HISTORY.md in the `Unreleased` section to describe your fix? Thanks.

PR #1563, opened 2017-05-04
Fix failed jobs page "argument out of range" error

streetlogics:
The failed_at date is always set to basically a UTC formatted date value (https://github.com/resque/resque/blob/master/lib/resque/failure/redis.rb#L17), but this line on the "failures" page on the Resque UI fails because the date format it's expecting for `Time.parse` is `YYYY-mm-dd HH:MM:SS Z` instead of `YYYY/mm/dd HH:MM:SS Z`.  DateTime.parse however properly handles the currently set format.  I guess an alternative would be to change the format that gets stored in redis in lib/resque/failure/redis.rb, but that wouldn't necessarily fix jobs already stored with the incorrect date format.

### DIFF
--- a/lib/resque/server.rb
+++ b/lib/resque/server.rb
@@ -102,6 +102,8 @@ module Resque
         Array(args).map do |a|
           a.to_yaml
         end.join("\n")
+      rescue
+        args.to_s
       end
 
       def worker_hosts


streetlogics:
Am trying to make sure this fix is included in the 1.X version of releases (1.27.5 maybe?), so created this branch off of the 1.27.4 tag.

Tonkpils:
@streetlogics can you branch off master and if necessary a backport would be made to the 1.x releases since that's the current stable

streetlogics:
@Tonkpils  - I went ahead and recreated this in https://github.com/resque/resque/pull/1566

PR #1568, opened 2017-06-06
Update Plugins.md to include right link to Wiki

lucas-ibotta:
:wave: hi!

I noticed this link has been moved and not updated since 2010.


chrisccerami:
Thanks!

Issue #1075, opened 2013-07-10
Shuffling queues on *


MichaelXavier:
Appologies if there is another issue about this but I couldn't find one. I found it an odd choice that when \* is specified, workers process queues in alphabetical order. I can't imagine most apps name queues where the name is both sensical and in the order that they want the work processed. Seems to be 2 different concerns being conflated there.

What we've seen in production is that the default workers end up starving other queues if they hit a queue full of hard work that takes precedence alphabetically.

I took a look at WorkerQueueList and MultiQueue and it seems like the best solution would be to have WorkerQueueList to shuffle for the \* queue and for MultiQueue to use a circular list when popping jobs in the non-blocking interface. If not, I think that if the first queue in a list has a high volume of quick jobs and the second queue has a low volume of slow jobs, the slow jobs will get starved.

Thoughts?

ermanc:
@steveklabnik I read the short discussion about this on http://librelist.com/browser//resque/2013/7/15/shuffling-queues-on/#4829b6fee392b1bca969781ba2462ac9

Though it doesn't seem like that went anywhere, there also doesn't seem to be any specific objections to this being at least a non-default option. 

We've been running into the issue of having one obnoxious queue blocking all other queues lexically below itself to be worked by our "general" workers (`QUEUE=*`).

Since the idea with `*` is that we don't necessarily care about the ordering (as you mentioned above) it might be possible to provide an environment variable flag to allow shuffling as part of glob pattern selection. 

This seems particularly simple, where all we might need is a simple `ENV['SHUFFLE']` check before `.sort`ing here: https://github.com/resque/resque/blob/master/lib/resque/worker.rb#L205, and do a `.shuffle` instead.

I will be getting a pull request ready, in case this sounds reasonable to you guys.

PR #1586, opened 2017-10-19
Shuffle glob queues per env flag

ermanc:
Allows developers the option to have shuffling `*` queue processing via `ENV['SHUFFLE']`.

The reasoning behind is two fold:

1. As mentioned by @steveklabnik in https://github.com/resque/resque/issues/1075#issuecomment-20768201, `*` processing kind of implies lack of explicit prioritization. If you really care about certain prioritization to be maintained, explicit queue lists are suited. 

2. Alphabetical prioritization means queues that are processed by general `*` workers are more prone to having one obnoxious queue (on purpose or accidentally) effectively block others just because they are higher up alphabetically. Using shuffling, an obnoxious queue cannot block or even slow down other queues as all have statistically similar chances of being processed every turn.

Usage is non-default, so the default behavior is still sorted `*` queues. 

Feature is enabled via `SHUFFLE=true QUEUE=* rake resque:work`

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -200,9 +200,15 @@ module Resque
     end
 
     def glob_match(pattern)
-      Resque.queues.select do |queue|
+      matches = Resque.queues.select do |queue|
         File.fnmatch?(pattern, queue)
-      end.sort
+      end
+
+      if ENV['SHUFFLE'] && ENV['SHUFFLE'] != 'false'
+        matches.shuffle
+      else
+        matches.sort
+      end
     end
 
     # This is the main workhorse method. Called on a Worker instance,
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -399,6 +399,33 @@ describe "Resque::Worker" do
     assert_equal %w( jobs high critical blahblah ).sort, processed_queues
   end
 
+  it "shuffles * queues if shuffling is set" do
+    # Since testing that something is shuffled and not sorted
+    # is impossible, we can only make false negatives less likely.
+    #
+    # 16 unique queue names when shuffled can only appear sorted
+    # 1 in 16! times (20,922,789,888,000 to be precise)
+    #
+    # Meaning we may certainly see this test fail if this test is run
+    # 10 times a day every day until the death of the sun.
+    new_queues = 16.times.map { |i| "queue_#{i}" }
+    new_queues.each { |q| Resque::Job.create(q.to_sym, GoodJob) }
+    all_queues = ['jobs'] + new_queues
+
+    processed_queues = []
+    @worker = Resque::Worker.new("*")
+    without_forking do
+      with_shuffling do
+        @worker.work(0) do |job|
+          processed_queues << job.queue
+        end
+      end
+    end
+
+    # Queues should not have been processed in sorted order
+    assert all_queues.sort != processed_queues
+  end
+
   it "works with globs" do
     Resque::Job.create(:critical, GoodJob)
     Resque::Job.create(:test_one, GoodJob)
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -286,3 +286,13 @@ def with_background
     ENV["BACKGROUND"] = old_background
   end
 end
+
+def with_shuffling
+  old_shuffle = ENV["SHUFFLE"]
+  begin
+    ENV["SHUFFLE"] = "true"
+    yield
+  ensure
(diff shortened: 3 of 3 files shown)


kbarrette (approved):

test/worker_test.rb
```diff
@@ -399,6 +399,33 @@ def self.perform
...
+      with_shuffling do
+        @worker.work(0) do |job|
+          processed_queues << job.queue
+        end
+      end
+    end
+
+    # Queues should not have been processed in sorted order
+    assert all_queues.sort != processed_queues
+  end
```
gerardc: This is cool. My only concern is that the test is quite complex. Since the guts of the change is using `shuffle` instead of `sort` how about trying to test `#glob_match` in isolation instead? Is that possible?

ermanc: It's possible but I personally prefer the current spec as it demonstrates the purpose and usage like the other specs here. It's also very loosely coupled to implementation.

gerardc:
Ok, I’m not sure I agree but fair enough. I think a unit test is enough to
cover what you’re doing; setting an ENV variable and then checking that a
method has been called. You could always demonstrate the functionality with
documentation in the README. Regardless the change is interesting :)

test/worker_test.rb
```diff
@@ -399,6 +399,33 @@ def self.perform
...
+    without_forking do
+      with_shuffling do
+        @worker.work(0) do |job|
+          processed_queues << job.queue
+        end
+      end
+    end
+
+    # Queues should not have been processed in sorted order
+    assert all_queues.sort != processed_queues
```
rafaelfranca: Even that it is small I think there is a chance that the shuffled array will be the same as the sorted array right? Would not this test be broken in that case?

lib/resque/worker.rb
```diff
@@ -200,9 +200,15 @@ def queues
     end
 
     def glob_match(pattern)
-      Resque.queues.select do |queue|
+      matches = Resque.queues.select do |queue|
         File.fnmatch?(pattern, queue)
-      end.sort
+      end
+
+      if ENV['SHUFFLE'] && ENV['SHUFFLE'] != 'false'
```
jeremy: Set up wildcard glob sort/shuffle behavior in `#initialize`

jeremy (commented):
The SHUFFLE env var suggests that the queue list is shuffled, or that all jobs are somehow shuffled. Can we be clear that it's shuffling the queues matched by the wildcard glob?

test/worker_test.rb
ermanc: There is a chance the shuffled array will be the same as the sorted array, yes. I'm more confident in a test case that tests the observable effect rather than implementation (e.g. spying on the `sort` vs `shuffle` call) and I cannot think of another way to test the "shuffled" outcome other than it being not sorted. 

I certainly welcome any suggestions on a better testing strategy, as I find this an interesting behavior to test for.

rafaelfranca: Yeah. I think the comment with the probability is good enough for that.

jrochkind:
i am interested in this, for reasons explained in PR description at top. 

Is there any hope for it? Anything I can do to help bring it home?

lib/resque/worker.rb
iloveitaly: What do think of making this a little more forgiving?

```suggestion
      if ENV['SHUFFLE'] && ENV['SHUFFLE'].downcase != 'false'
```

iloveitaly (changes requested):
This is a _very_ old PR, but still a great change! Thanks for the contribution. I've slowly been going through old PRs since I've been added as a contributor.

@ermanc if there's still interest in getting this merged, could you:

* Rebase on master and make sure tests are passing
* Add some docs to the readme about how this works

Also, if someone else wants to pick this up (@jrochkind) that's great too!

PatrickTulskie:
Thanks for this, and sorry for the long silence.

Closing as stale bc `glob_match` has been resignatured since this was opened (it now takes the queue list as an argument and handles skip_queues as well), so the patch no longer applies and the surrounding logic has moved.

The idea still has merit: always draining wildcard-matched queues in alphabetical order isn't great for fairness. If we add it, I'd want it as a real configuration option rather than an ENV['SHUFFLE'] global so it composes with the rest of the worker config. Happy to revisit it if we can make those changes. Frankly, it's probably easier to make a new PR that references this one though than try to rebase this against Resque 3+

PR #1594, opened 2017-12-07
Prevent thundering herd when pruning dead workers

fw42:
We've recently had a small service disruption where a large number of workers died and were dead for longer than `Resque.prune_interval`. Once the problem was fixed, all of our (thousands of) workers were able to boot again around the same time, meaning that all of them tried to prune all of the dead old workers at once (see [Thundering herd problem](https://en.wikipedia.org/wiki/Thundering_herd_problem)), therefore causing a huge load spike (pruning workers is a somewhat expensive operation, involving multiple server roundtrips to Redis).

This PR fixes that by making it so that only one worker can be pruning and all subsequent attempts will be skipped for the next `Resque.prune_interval` seconds.

@es @pushrax @xldenis @hkdsun @sirupsen
cc @chrisccerami

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -44,6 +44,7 @@ module Resque
                               :heartbeat!,
                               :remove_heartbeat,
                               :all_heartbeats,
+                              :acquire_pruning_dead_worker_lock,
                               :set_worker_payload,
                               :worker_start_time,
                               :worker_done_working
@@ -275,6 +276,10 @@ module Resque
         @redis.hgetall(HEARTBEAT_KEY)
       end
 
+      def acquire_pruning_dead_worker_lock(worker, expiry)
+        @redis.set(redis_key_for_worker_pruning, worker.to_s, :ex => expiry, :nx => true)
+      end
+
       def set_worker_payload(worker, data)
         @redis.set(redis_key_for_worker(worker), data)
       end
@@ -298,6 +303,10 @@ module Resque
 
       def redis_key_for_worker_start_time(worker)
         "#{redis_key_for_worker(worker)}:started"
+      end
+
+      def redis_key_for_worker_pruning
+        "pruning_dead_workers_in_progress"
       end
     end
 
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -669,6 +669,42 @@ describe "Resque::Worker" do
     assert_equal [], Resque::Worker.all_workers_with_expired_heartbeats
   end
 
+  it "does not prune if another worker has pruned (started pruning) recently" do
+    now = Time.now
+    workerA = Resque::Worker.new(:jobs)
+    workerA.to_s = 'workerA:1:jobs'
+    workerA.register_worker
+    workerA.heartbeat!(now - Resque.prune_interval - 1)
+    assert_equal 1, Resque.workers.size
+    assert_equal [workerA], Resque::Worker.all_workers_with_expired_heartbeats
+
+    workerB = Resque::Worker.new(:jobs)
+    workerB.to_s = 'workerB:1:jobs'
+    workerB.register_worker
+    workerB.heartbeat!(now)
+    assert_equal 2, Resque.workers.size
+
+    workerB.prune_dead_workers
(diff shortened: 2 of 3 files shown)


lib/resque/data_store.rb
```diff
@@ -275,6 +276,10 @@ def all_heartbeats
         @redis.hgetall(HEARTBEAT_KEY)
       end
 
+      def acquire_pruning_dead_worker_lock(worker, expiry)
+        @redis.set(redis_key_for_worker_pruning, worker.to_s, :ex => expiry, :nx => true)
```
fw42: I just realized that this command is only available in Redis version >= 2.6.12 and `redis-rb` version >= 3.05. Not sure what to do about that. It's possible to reimplement the same behaviour via a combination of `MULTI`, `SETNX`, `EXPIRE`, and `WATCH`, but that feels kinda hacky. Any suggestions? Could just disable this feature completely if the gem or the server version are too old. Thoughts?

fw42: Could also do something slightly different:
```ruby
def acquire_pruning_dead_worker_lock(worker, expiry)
  setnx, expire = @redis.multi do
    @redis.setnx(redis_key_for_worker_pruning, worker.to_s)
    @redis.expire(redis_key_for_worker_pruning, expiry)
  end

  setnx
end
```
but the semantics are slightly different since this version bumps the expiry on every call while the other version only sets the expiry ones and then never bumps it until expires. The second version could lead to the key never expiring if workers reboot frequently (frequent deploys).

sirupsen: Is it easy to get the version? I'm somewhat worried about 1000s of workers asking the server for its version on deploy.

sirupsen: In the README it says:

```
Resque now supports Ruby 2.0.0 and above. Any future updates will not be guaranteed to work without defects on any Rubies older than 2.0.0. We will also only be supporting Redis 3.0 and above going forward.
```

fw42: Yea good point. The Redis `INFO` command returns the server version. But I'm not sure how expensive that command is (probably much less expensive than doing the entire pruning though, but the concern remains).

sirupsen: Support for old Redis was removed in this PR https://github.com/resque/resque/pull/1594

sirupsen: Since that PR we only support `>= 2.6` [where `TIME` was introduced](https://redis.io/commands/time). 

I'd say you're good here, @fw42

hkdsun:
What's the rationale for re-using `Resque.prune_interval` as the lock expiry? That wouldn't allow cleanup of workers for 5 minutes by default, which seems too long to me. There's no right value that I can suggest but maybe something closer to the heartbeat interval makes more sense?

fw42:
Yea I don't really have a strong opinion on the expiry time. I thought 5 minutes makes sense since that is about the time it takes for us to deploy and restart all our Resque workers within one region. If we set it to the heartbeat interval (1 minute) then each deploy might cause multiple prunings, which seems unnecessary.

hkdsun:
If the intention is to prevent a herd effect even a short interval would do the job, regardless of the number of times the pruning is done. I say we go with something like 10 seconds unless we have a good reason not to

fw42:
Changed the expiry from `Resque.prune_interval` (default 5 minutes) to `Resque.heartbeat_interval` (default 60 seconds).

test/worker_test.rb
```diff
@@ -669,6 +669,42 @@ def self.perform
...
+    workerD = Resque::Worker.new(:jobs)
+    workerD.to_s = 'workerD:1:jobs'
+    workerD.register_worker
+    workerD.heartbeat!(now)
+    assert_equal 3, Resque.workers.size
+
+    # workerC does not get pruned because workerB already pruned recently
+    workerD.prune_dead_workers
+    assert_equal [workerC], Resque::Worker.all_workers_with_expired_heartbeats
+  end
```
pushrax: Ideally this would also ensure that the lock does ever expire, this could perhaps be done by stubbing `Resque.heartbeat_interval`

fw42:
@chrisccerami: Could you release a new gem version if you have a minute? Thanks!

PR #1602, opened 2018-02-01
Stop unconditionally require-ing the Airbrake gem for Resque::Failure::Airbrake

unleashed:
Check first for an `::Airbrake` constant if defined, so that there's no unconditional dependency on the `airbrake` gem.

This makes this module usable with other implementations of the same interface.

While at it also fix the test and some relative references to the constant.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,9 @@
 
 ### Changed
 * Remove support for Rubies < 2.0
+* Do not require the Airbrake gem unconditionally anymore when using the
+  Resque::Failure::Airbrake module. A top-level Airbrake constant implementing
+  the same interface will be used if present.
 
 ## 1.27.4 (2017-04-15)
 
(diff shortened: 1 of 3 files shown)


rafaelfranca:
I'm not sure if this is a good idea. For example, why only the `Airbreak` notifier will support this? Do you think we should allow other people to define the same constant name as a gem and our code just work with it?

It is not only looks wrong but also is dangerous. It means that a malicius gem can now ship a airbreak module that have the same API of the gem and we will execute the methods thinking it is the airbreak gem.

I think it is better to you to define your own adapter. Thank you for the pull request.

rafaelfranca:
Right now I'm more inclined to remove those adapter from resque and let their evolve along with the onwers of the adapter like suggested by @jeremy here https://github.com/resque/resque/pull/1511#pullrequestreview-95976147.

unleashed:
In the specific case of Airbrake IIRC that means there will be no support for versions lower than 5, which is maybe 2 years old. Other than that the decision is probably more about whose responsibility is to integrate with the other and how much maintenance that code could realistically see.

I think that removing these adapters altogether, or leaving them functional with some form of warning to have people check their gems for best support, might be the wiser decision.

PR #1608, opened 2018-02-15
Remove Ruby warnings related to hooks

rafaelfranca:
Initialize the @hooks to avoid warnings and also some conditionals and memoization.

I also delete the unused definition of `hooks` that was being redefined by the `hooks(name)` method below and never was used.

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -561,11 +561,6 @@ module Resque
     @hooks[name] = []
   end
 
-  # Retrieve all hooks
-  def hooks
-    @hooks
-  end
-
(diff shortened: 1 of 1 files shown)


lib/resque.rb
```diff
@@ -543,33 +543,27 @@ def sample_queues(sample_size = 1000)
 
   private
 
+  @hooks = Hash.new([])
```
jeremy: Needs `Hash.new { |h, k| h[k] = [] }` so `@hooks[name].concat(block)` works?

PR #1612, opened 2018-02-27
Fix Bug in Failure::RedisMultiQueue

atareshawty:
This PR introduces a failing test that exposed a bug in `RedisMultiQueue.each`. My team utilizes the Resque UI to look at job failures for bug triaging. We recently switched our Failure backend from `Redis` to `RedisMultiQueue`. Soon after, it seemed that the jobs we were attempting to remove through the UI weren't the ones actually getting removed. After some debugging, the conclusion was that `RedisMultiQueue.each` doesn't work as expected. 

We're using version `1.27.4`.

The first commit introduces some tests around the method including one that breaks. The issue is calling `.each` for a particular queue with an offset > 0 in descending order. The indices yielded don't point to the correct jobs in the failure queue. This is why the UI was restarting jobs different than the ones clicked.

If you agree that this is a bug, I can go ahead and make the fix. The correct implementation already exists in `Failure::Redis` so it should be a simple change. Hopefully this is enough information. If you need anything else, let me know! 😄

### DIFF
--- a/lib/resque/failure/redis_multi_queue.rb
+++ b/lib/resque/failure/redis_multi_queue.rb
@@ -60,7 +60,7 @@ module Resque
         end
         items.each_with_index do |item, i|
           if !class_name || (item['payload'] && item['payload']['class'] == class_name)
-            id = reversed ? (items.length - 1) - (offset + i) : offset + i
+            id = reversed ? (items.length - 1) + (offset - i) : offset + i
             yield id, item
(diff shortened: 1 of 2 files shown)


atareshawty:
Ready for 👀!

jeremy:
Thanks @atareshawty!

gerardc:
Nice work!

PR #1616, opened 2018-03-17
Fixing README.markdown

akshitahluwalia:
Fixing a logical bug in documentation code.


README.markdown
```diff
@@ -225,7 +225,7 @@ Resque workers are rake tasks that run forever. They basically do this:
 ``` ruby
 start
 loop do
-  if job = reserve
+  if job == reserve
```
jeremy: This is intended to be an assignment 😊

PR #1620, opened 2018-04-14
Adding docs on heartbeat_internal, prune_interval, and enqueue_front

iloveitaly:
Title is self-explanatory. I couldn't find any docs on these options, so I figured some basic comments would be a good place to start.


jeremy:
Thanks @iloveitaly!

PR #1624, opened 2018-04-16
Report a job that pruned worker was processing

kirs:
When you run into `PruneDeadWorkerDirtyExit` in production, you have no idea what job the pruned worker was processing. That's what the PR is trying to address. It also makes the exception message a bit more user friendly.

@hkdsun @fw42 this should help us to find out if there's any pattern in workers that are killed in ungraceful way.

### DIFF
--- a/lib/resque/errors.rb
+++ b/lib/resque/errors.rb
@@ -14,7 +14,13 @@ module Resque
       super message
     end
   end
-  class PruneDeadWorkerDirtyExit < DirtyExit; end
+
+  class PruneDeadWorkerDirtyExit < DirtyExit
+    def initialize(hostname, job)
+      job ||= "<Unknown Job>"
+      super("Worker #{hostname} did not gracefully exit while processing #{job}")
+    end
+  end
 
   # Raised when child process is TERM'd so job can rescue this to do shutdown work.
   class TermException < SignalException; end
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -610,7 +610,9 @@ module Resque
         # client library or an older version of Resque. We won't touch these.
         if all_workers_with_expired_heartbeats.include?(worker)
           log_with_severity :info, "Pruning dead worker: #{worker}"
-          worker.unregister_worker(PruneDeadWorkerDirtyExit.new(worker.to_s))
+
+          job_class = worker.job(false)['payload']['class'] rescue nil
+          worker.unregister_worker(PruneDeadWorkerDirtyExit.new(worker.to_s, job_class))
           next
         end
 
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -776,6 +776,23 @@ describe "Resque::Worker" do
     end
   end
 
+  it "correctly reports a job that the pruned worker was processing" do
+    workerA = Resque::Worker.new(:jobs)
+    workerA.to_s = "jobs01.company.com:3:jobs"
+    workerA.register_worker
+
+    job = Resque::Job.new(:jobs, {'class' => 'GoodJob', 'args' => "blah"})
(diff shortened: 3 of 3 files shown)


lib/resque/errors.rb
```diff
@@ -14,7 +14,12 @@ def initialize(message=nil, process_status=nil)
       super message
     end
   end
-  class PruneDeadWorkerDirtyExit < DirtyExit; end
+
+  class PruneDeadWorkerDirtyExit < RuntimeError
```
jeremy: Reasoning for changing the superclass? (Breaks compatibility)

lib/resque/worker.rb
```diff
@@ -610,7 +610,14 @@ def prune_dead_workers
         # client library or an older version of Resque. We won't touch these.
         if all_workers_with_expired_heartbeats.include?(worker)
           log_with_severity :info, "Pruning dead worker: #{worker}"
-          worker.unregister_worker(PruneDeadWorkerDirtyExit.new(worker.to_s))
+
+          job_class = begin
+            worker.job(false)['payload']['class']
+          rescue => _error
+            "<Unknown Job>"
```
jeremy: Since `PruneDeadWorkerDirtyExit` is responsible for creating the exception message, would set `job_class` to `nil` here and let `PruneDeadWorkerDirtyExit#initialize` use e.g. `job ||= "<Unknown Job>"`

lib/resque/worker.rb
```diff
@@ -610,7 +610,14 @@ def prune_dead_workers
         # client library or an older version of Resque. We won't touch these.
         if all_workers_with_expired_heartbeats.include?(worker)
           log_with_severity :info, "Pruning dead worker: #{worker}"
-          worker.unregister_worker(PruneDeadWorkerDirtyExit.new(worker.to_s))
+
+          job_class = begin
+            worker.job(false)['payload']['class']
+          rescue => _error
```
jeremy: Can omit the unused ` => _error`.

lib/resque/errors.rb
kirs: I changed it because `PruneDeadWorkerDirtyExit` would now have different arguments than `DirtyExit`.

It's arguable that `PruneDeadWorkerDirtyExit` is a meta-exception that is not something that you would `rescue`, but I agree that it's better to keep the original superclass.

kirs:
Thanks for a prompt review @jeremy! I updated the PR.

lib/resque/worker.rb
```diff
@@ -610,7 +610,14 @@ def prune_dead_workers
         # client library or an older version of Resque. We won't touch these.
         if all_workers_with_expired_heartbeats.include?(worker)
           log_with_severity :info, "Pruning dead worker: #{worker}"
-          worker.unregister_worker(PruneDeadWorkerDirtyExit.new(worker.to_s))
+
+          job_class = begin
+            worker.job(false)['payload']['class']
+          rescue
+            nil
+          end
```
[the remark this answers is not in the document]
kirs: I tried to avoid that because it’s discouraged by bbatsov/ruby-style-guide, but can change since Resque itself has no guidline.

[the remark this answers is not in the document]
kirs: Updated.

ravindra9278:
Hi

Today I got same error, I am not sure where is the issue.

```
Worker f9635fe4-ccec-4350-9b48-f209f4705d08:4 on INVOICES at 14 minutes ago Retry or Remove
Class Restaurant::Invoicer::RunJob
Arguments --- '1162'
Exception Resque::PruneDeadWorkerDirtyExit
Error f9635fe4-ccec-4350-9b48-f209f4705d08:4:invoices
```

PR #1627, opened 2018-05-01
Support secure Redis URLs

### LINKED ISSUE #1626: Support Redis library's SSL abilities
https://github.com/resque/resque/blob/eadb02a7a3afb63fe926cd59e3e943a1b3dab547/lib/resque.rb#L115

Any chance we can make this a little smarter so it detects rediss:// lines and enables SSL on the connection?  Maybe `/rediss?\:\/\//` instead?  Thanks

electrickite:
Resolves #1626 by extending the regex in `redis=` to allow `rediss://` secure URLs in addition to the standard `redis://`

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -112,7 +112,7 @@ module Resque
   def redis=(server)
     case server
     when String
-      if server =~ /redis\:\/\//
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thanks

PR #1629, opened 2018-05-07
Fix Airbrake failure backend

Alfarlost:
Airbrake failure backend doesn't work now. And the tests are skipped because no airbrake dependency in Gemfile


jeremy:
Thanks @Alfarlost!

PR #1631, opened 2018-05-29
Resque::Failure::Multiple calls `save` to all backends even if some backends fail

labocho:
When some failure backend raise error during `Resque::Failure::Multiple#save`, the rest backends' `save` are never called.

```ruby
class BrokenBackend < Resque::Failure::Base
  def save
    raise "Broken"
  end
end

Resque::Failure::Multiple.classes = [BrokenBackend, Resque::Failure::Redis]
Resque::Failure.backend = Resque::Failure::Multiple

# When a job fails, Resque::Failure::Redis#save is never called.
```

I want to call all backends' `save` for redundancy. I wrote patch to call all backends and wrap original error(s).

### DIFF
--- a/lib/resque/failure/multiple.rb
+++ b/lib/resque/failure/multiple.rb
@@ -33,9 +33,11 @@ module Resque
         errors = {}
 
         @backends.each do |backend|
-          backend.save
-        rescue
-          errors[backend.class] = $!
+          begin
+            backend.save
+          rescue
+            errors[backend.class] = $!
+          end
         end
 
         unless errors.empty?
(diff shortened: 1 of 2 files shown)


matthewhively:
This would be cool if this could be fixed up and merged in.

PatrickTulskie:
@labocho hey I know it's been like... 8 years since you opened this PR, but I think we can probably get it into the next minor release if you rebase your PR on the latest so we can clear out the merge conflict.

PatrickTulskie:
We pulled this change into #1964, making sure to attribute the work to @labocho and then also addressed some issues that related to recent failure backend changes. I'm going to close this out as superseded by #1964

PR #1636, opened 2018-06-15
Use Minitest 5 for test.

junaruga:
This PR enables to use minitest 5.
I updated it to the latest version.
https://rubygems.org/gems/minitest

As "assert_equal nil, foo" is deprecated from Minitest 5, I fixed it.

```
DEPRECATED: Use assert_nil if expecting nil from /home/jaruga/git/resque/test/resque_test.rb:205. This will fail in Minitest 6.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ gemspec
 
 gem "json"
 
-gem "minitest", "4.7.0"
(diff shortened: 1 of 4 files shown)


junaruga:
ping.

PR #1638, opened 2018-06-18
Improve failed queues list on /queues UI

tsevan:
Explicitily check if `RedisMultiQueue` backend is enabled in `failed_multiple_queues?` helper.

With the current implementation in case if `RedisMultiQueue` failure backend is set and there is only one failed queue in the resque it would make an [assumption](https://github.com/resque/resque/blob/v1.27.4/lib/resque/server/views/queues.erb#L51-L54) the name of the queue is `failed`. When developer visits the `failed` link he or she is unable to see the failed jobs.

### DIFF
--- a/lib/resque/server/helpers.rb
+++ b/lib/resque/server/helpers.rb
@@ -9,7 +9,9 @@ Resque::Server.helpers do
 
   def failed_multiple_queues?
     return @multiple_failed_queues if defined?(@multiple_failed_queues)
-    @multiple_failed_queues = Resque::Failure.queues.size > 1
+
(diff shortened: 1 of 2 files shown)


iloveitaly (approved):
@chrisccerami this looks like another great candidate for merging.

iloveitaly:
@chrisccerami what do you think about merging this PR in?

iloveitaly:
@tsevan great change! Could you submit another PR with a changelog entry?

tsevan:
Thanks for the review and the delivery @iloveitaly ! Here you go: https://github.com/resque/resque/pull/1766.

PR #1640, opened 2018-06-21
Update read me until the new version comes out

basicBrogrammer:
I'm not sure if this is needed with the future version. 
It looks like the fix is in master just not on 1.27.4. https://github.com/resque/resque/blob/v1.27.4/lib/resque.rb#L116
I'm not sure how y'all want to document this or when the next release is planning on going out. 
I just ran into this problem with resque 1.27.4 and redis >4.0. 
This change fixed it for me.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -801,7 +801,7 @@ rails_env = ENV['RAILS_ENV'] || 'development'
 config_file = rails_root + '/config/resque.yml'
 
 resque_config = YAML::load(ERB.new(IO.read(config_file)).result)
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Thank you for the pull request. The master branch documents the code in the master branch, so changing the documentation to workaround an issue that is already fixed doesn't seems right. If we need to document something that is fixed in master but broken in 1.27.4 it should be in the 1.27.4 readme, but as it is already released and we can't change there is no way to document.

PR #1652, opened 2018-10-30
allow to configure stat datastore

GustavoCaso:
This would allow developers to define there own `StatDataStore` 

So depending on your needs, you could use a different `redis` or simply create a null op


GustavoCaso:
@jeremy are we talking about removing the `def_delegators` from `data_store.rb` ?

jeremy:
@GustavoCaso that's what stood out to me, yes. Any other API changes to account for?

lib/resque/stat_data_store.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    def stat(stat)
+      redis.get("stat:#{stat}").to_i
+    end
+
+    def increment_stat(stat, by = 1)
+      redis.incrby("stat:#{stat}", by)
+    end
+
+    def decremet_stat(stat, by = 1)
```
moechaieb: Nitpick: this is a good opportunity to fix this typo :)

rafaelfranca:
I tihink it is possible to implement this without any backwards incompatible change and I'd prefer to do that.

For example, we don't need to change `Resque::Stat` from module to a class. We can keep it as a module and have a module accessor for the datastore and delegate `Resque.stat_data_store=` to `Resque::Stat.data_store =`.

Same thing with the `Resque::DataStore` we can keep that class untouched.

lib/resque.rb
```diff
@@ -147,6 +147,15 @@ def redis_id
     data_store.identifier
   end
 
+  def stat_data_store=(stat_data_store)
+    stat = Resque::Stat.data_store = stat_data_store
+    @stat_data_store = stat
+  end
+
+  def stat_data_store
+    @stat_data_store ||= Resque::Stat
```
rafaelfranca: The return of this method should be `Resque::Stat.data_store` otherwise you will get either an object that respond to those 4 methods we know or a object that respond to all methods `Resque::Stat` respond to. It will be inconsistent.

lib/resque/stat.rb
```diff
@@ -7,12 +7,14 @@ module Resque
...
-      Resque.redis
+
+    def data_store
+      @data_store ||= Resque.redis
+    end
+
+    def data_store=(data_store)
+      @data_store = data_store
     end
-    alias :data_store :redis
```
rafaelfranca: Can we keep the alias to `redis` with a deprecation warning?

test/resque_test.rb
```diff
@@ -287,6 +287,12 @@
   end
 
   describe "stats" do
+    it "allows to set custom stat_data_store" do
+      dummy = DummyStatStore.new
+      Resque.stat_data_store = dummy
```
rafaelfranca: You need to store the previous data_store and reset in an ensure block otherwise tests will fail.

GustavoCaso: Done https://github.com/resque/resque/pull/1652/files#diff-8ce9de84c5e22ef89d9f1e236811c946R6

lib/resque/stat.rb
GustavoCaso: Done https://github.com/resque/resque/pull/1652/commits/7f79b5b17d275fe664f747dc2ddb31994e66b056

moechaieb:
@jeremy @rafaelfranca we'd love to get some 👀 on this when you get the chance :)

lib/resque/stat.rb
```diff
@@ -7,12 +7,19 @@ module Resque
...
-    # Direct access to the Redis instance.
+
     def redis
-      Resque.redis
+      warn '[Resque] [Deprecation] Resque::Stat #redis method is deprecated (please use #data_strore)'
+      data_store
+    end
+
+    def data_store
+      @data_store ||= Resque.redis
```
fw42: This caching is a bit useless (no performance gain) and seems even dangerous or at least counterintuitive. If I change `Resque.redis` to point to somewhere else, I would expect things to work correctly, but since this module here now has the "old" value memoized, it would continue using the old one.

PR #1656, opened 2018-11-13
Add new backend that does not allow duplicates

sadhu89:
With the `Resque::Failure::Redis` class, a job failure instance keeps being added to redis until someone fixes it. So sometimes, we end up with 500 failed jobs in the dashboard where the actual number of unique failures is 15. (This happens often for poller jobs)

This PR introduces a class `Resque::Failure::RedisUniqueFailures` that stops resque from adding multiple instances of the same failed jobs.

- It also fixes the tests for the other backends that were representing incorrectly the `payload` as including `"class" => Object` instead of `"class" => "Object"`

### DIFF
--- a/lib/resque/failure/redis_unique_failures.rb
+++ b/lib/resque/failure/redis_unique_failures.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'resque/failure/redis'
+
+module Resque
+  module Failure
+    class RedisUniqueFailures < Redis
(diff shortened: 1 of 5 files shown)


rafaelfranca:
Thank you so much for the pull request. While I understand the use case I don't think this require a new default backend on this gem. It is easy to define backends in applications to fulfill the needs and configure Resque to use this new backend.

I'm going to apply the last commit, though.

PR #1657, opened 2018-11-20
Add .all_localhost method

kyoshidajp:
Add `.all_localhost` method to `Resque::Worker`.

For example, to stop the worker process gracefully in multi-host, since `Resque::Worker.all` returns all worker objects, it does not satisfy the requirement. This method returns ONLY their workers.


rafaelfranca:
Can you expand better why you need this method?

kyoshidajp:
Sorry for late reply.
This is my mistake. I can access without this method by another way. So, I close this. Thanks.

PR #1659, opened 2018-11-26
Add `requeue_queue` method to `Resque::Failure::Multiple`

jzaleski:
This change adds the `requeue_queue` method to `Resque::Failure::Multiple` this is a hard requirement for the "retry-all" functionality to work via the Web UI and is currently causing issues w/in the `resque-retry` gem (which I maintain). This also seems to be an innocuous change to `Resque` itself.

### DIFF
--- a/lib/resque/failure/multiple.rb
+++ b/lib/resque/failure/multiple.rb
@@ -60,6 +60,10 @@ module Resque
         classes.first.requeue_all
       end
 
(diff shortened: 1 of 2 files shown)


jzaleski:
I just added a test that covers the related functionality. Please let me know if there is anything else you need. Thanks @jeremy !

PR #1662, opened 2019-01-06
Remove leading white spaces in a blank line

yoshino:
I removed white spaces.

screenshot from 2019-01-06 23-11-31


jeremy:
Thanks for the PR, @yoshino! We generally don't apply cosmetic changes like whitespace correction since they're noisy in `git blame`. We can add a basic Rubocop config for whitespace checking, however. I'll start a PR and pull in your commit!

PR #1668, opened 2019-01-22
Switch queue direction in preperation for resilient queues

BoGs:
This is the first PR for resilient queues using `rpoplpush` within Resque. This PR changes the direction of the jobs pushed to the queue and the location where the jobs are poped off the queue. This is required since there is no `lpoprpush` command in Redis. 

I will follow up with another PR to make the changes required for utilizing `rpoplpush`.

### DIFF
--- a/test/resque_test.rb
+++ b/test/resque_test.rb
@@ -219,17 +219,17 @@ describe "Resque" do
     end
 
     it "can peek at a queue" do
-      assert_equal({ 'name' => 'chris' }, Resque.peek(:people))
+      assert_equal({ 'name' => 'mark' }, Resque.peek(:people))
       assert_equal 3, Resque.size(:people)
     end
 
     it "can peek multiple items on a queue" do
       assert_equal({ 'name' => 'bob' }, Resque.peek(:people, 1, 1))
 
-      assert_equal([{ 'name' => 'bob' }, { 'name' => 'mark' }], Resque.peek(:people, 1, 2))
-      assert_equal([{ 'name' => 'chris' }, { 'name' => 'bob' }], Resque.peek(:people, 0, 2))
-      assert_equal([{ 'name' => 'chris' }, { 'name' => 'bob' }, { 'name' => 'mark' }], Resque.peek(:people, 0, 3))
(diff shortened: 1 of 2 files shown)


rafaelfranca:
How deploying this change would affect an application? I guess it would change the execution order of the jobs that were enqueued before the deploy started right? Should we care about that?

test/resque_test.rb
```diff
@@ -219,17 +219,17 @@
     end
 
     it "can peek at a queue" do
-      assert_equal({ 'name' => 'chris' }, Resque.peek(:people))
+      assert_equal({ 'name' => 'mark' }, Resque.peek(:people))
```
rafaelfranca: Now the behavior of `peek` has changed. While before we would get the next job that would be executed now we would get the last one.

jeremy:
Redis issue for LPOPRPUSH: https://github.com/antirez/redis/issues/658

Note that LPOPRPUSH is [available in rxlists](https://github.com/RedisLabsModules/redex#lpoprpush-srclist-dstlist).

Seems that using RPOPLPUSH to move from queue list to working list, we'd still want LPOPRPUSH to move back from working list to queue list?

Alternatively, we could leave the queue in the same order and implement our operations in Lua instead: https://github.com/antirez/redis/issues/658#issuecomment-423190313

taf2:
question why not include a lua script to define lpoprpush - i think the underlying redis version is already high enough to require lua scripting?

a simple lpoprpush.lua script:
```
local src_list = ARGV[1]
local dst_list = ARGV[2]
local value = redis.call('lpop', src_list)
if value then -- avoid pushing nils
  redis.call('rpush', dst_list, value)
end
return value
```

This would not require a direction change in the queue... and might make it easier to migrate?

taf2:
Few other thoughts to dump here... if upgrading to this the list would not be lost so no data is lost however any jobs that where dependent on order would not be processed in order (maybe matters maybe not) - so say you upgraded and were lucky enough to have a slow time in your app, upgrading the might be the only way to mitigate.  Ideally an upgrade would transition from one active redis/resque cluster to another such that the order of jobs processed would not be impacted.  Other considerations for this is how existing plugins might be impacted do plugins exist today that make assumptions around the use of  rpush instead of lpush?

PR #1673, opened 2019-02-28
fix plugins link

TheInan:


jeremywadsack:
Appears to have been fixed in #1672.

PR #1676, opened 2019-03-07
add compatibility warning to readme

### LINKED ISSUE #1675: segfaulting with 1.27.4+ (including 2.0) and ruby 2.5.3
After upgrading from 1.26.0 to 1.27.4, we started seeing segfaults when calling `ConditionVariable#signal`, introduced in https://github.com/resque/resque/pull/1484.

Looks like [this is a known bug in Ruby 2.5](https://bugs.ruby-lang.org/issues/14659). Perhaps it should be publicized as a warning in the readme.

[Status of ruby 2.5 fix looks like it's being tracked here.](https://bugs.ruby-lang.org/issues/14852)

michaelglass:
should resolve #1675

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -36,6 +36,9 @@ general usage stats, and helps you track failures.
 Resque now supports Ruby 2.3.0 and above.
 We will also only be supporting Redis 3.0 and above going forward.
 
+### Ruby 2.5.* Compatibility Warning
+Ruby 2.5.* has [a bug that affects Resque 1.27.0+](https://github.com/resque/resque/issues/1675). Consider using 1.26.0 or upgrade / downgrade your Ruby version.
+
 
 The Blog Post
 -------------


README.markdown
```diff
@@ -36,6 +36,9 @@ general usage stats, and helps you track failures.
 Resque now supports Ruby 2.3.0 and above.
 We will also only be supporting Redis 3.0 and above going forward.
 
+### Ruby 2.5.* Compatibility Warning
+Ruby 2.5.* has [a bug that affects Resque 1.2.7+](https://github.com/resque/resque/issues/1675). Consider using 1.2.6 or upgrade / downgrade your Ruby version.
```
AMHOL: I think this should be `1.26.0` not `1.2.6`?

michaelglass: fixed

rafaelfranca:
Resque is now 2.0, is this still relevant?

michaelglass:
@rafaelfranca yup, updated the issue to link to the [buggy-in-ruby-2.5 code that still exists in 2.0](https://github.com/resque/resque/blob/adb633a0f6b98b1eb5a5a85bb36ebac9309978fd/lib/resque/thread_signal.rb#L11)

README.markdown
```diff
@@ -36,6 +36,9 @@ general usage stats, and helps you track failures.
 Resque now supports Ruby 2.3.0 and above.
 We will also only be supporting Redis 3.0 and above going forward.
 
+### Ruby 2.5.* Compatibility Warning
+Ruby 2.5.* has [a bug that affects Resque 1.27.0+](https://github.com/resque/resque/issues/1675). Consider using 1.26.0 or upgrade / downgrade your Ruby version.
```
rafaelfranca: How about Ruby 2.6? Does it also have the problem?

rafaelfranca:
This issue is very strange because we do use Resque 2.0 with Ruby 2.5 at Shopify and we had no problem so far.

michaelglass:
I don't know enough about Resque but maybe there's a single-threaded mode?

iloveitaly:
It looks like the [this was fixed](https://github.com/ruby/ruby/commit/63d9ab33fe419167382fc03be2c00413a78c05cb) in ruby 2.5.4. Going to close this out, it's safe to assume users are going to use the latest minor patch version of ruby.

Issue #1647, opened 2018-09-18
Trigger events with ActiveSupport::Notifications when available


jeremywadsack:
We have several use cases were we'd like to perform actions when the Resque worker gets to a particular state:

1. The `resque-kubernetes` gem currently [monkey patches `Resque::Worker`](https://github.com/keylimetoolbox/resque-kubernetes/blob/v0.9.0/lib/resque/kubernetes/worker.rb) to terminate the worker when the queue is empty
2. Some use cases of this might require [additional handling](keylimetoolbox/resque-kubernetes/issues/16) when the worker terminates
3. I want to launch a clean-up job when a worker becomes idle

Case 3, and perhaps case 2, could be managed (without monkey patching) if we could add code that responded to an event. While case 1 might be useful as a first-class feature (it's switched on by an environment variable), we could rebuild it as an event listener that was less coupled. [Active Support Instrumentation](https://guides.rubyonrails.org/active_support_instrumentation.html) provides a nice framework for this; at least in cases where `ActiveSupport` is available.

Would you be open to a pull request that added a couple event notifications? The ones I'm thinking of are:
- State changed (or a separate notification when changed to "processing" and when changed to "idle")
- Worker terminated

I'm also open to using a different event framework that doesn't depend on `ActiveSupport` so that the event listeners can be implemented in non-Rails environments.

jeremywadsack:
While I've never seen it in use, it appears that Ruby has a [built-in `Observable`](http://ruby-doc.org/stdlib-2.5.1/libdoc/observer/rdoc/Observable.html), which might be the right choice here.

rafaelfranca:
I'm open to the idea. Can you open a PR with an initial implementation to show all the events that you think would be useful to observe?

jeremy:
Seen https://github.com/rwdaigle/resque_instrumentation ? It may fit the bill. It uses Resque's existing hooks as instrumentation points to fire AS Notifications.

jeremywadsack:
@jeremy Thank you for the link. I'm ashamed I didn't think to look at the plugins list where this is listed prominently! I'd love to use an existing plug-in and not have to add more to `resque`.

The use cases I'm looking at are two events: when the worker is "idle" (by which I mean there are no longer any jobs in the queue), and when the worker terminates. If you don't mind let me run some thoughts by you in case I'm missing something.

[...]

PR #1682, opened 2019-05-23
Add queue_empty and worker_exit hooks on the worker

jeremywadsack:
As discussed in #1647, to allow plugins and such to act on events such as the worker becoming idle (queue is now empty) and the worker exiting, this PR adds support for two new worker hooks.

  - `queue_empty` when the job queue empties and the worker becomes idle
  - `worker_exit` when the worker exits

Please review and let me know comments, thoughts. I'm happy to rename those if they don't feel "hook-y" enough. I wasn't sure, but also considered "after_queue_empty" and "before_exit".

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -148,6 +148,8 @@ module Resque
       @heartbeat_thread = nil
       @heartbeat_thread_signal = nil
 
+      @last_state = :idle
+
       verbose_value = ENV['LOGGING'] || ENV['VERBOSE']
       self.verbose = verbose_value if verbose_value
       self.very_verbose = ENV['VVERBOSE'] if ENV['VVERBOSE']
@@ -236,6 +238,7 @@ module Resque
         break if shutdown?
 
         unless work_one_job(&block)
+          state_change
           break if interval.zero?
           log_with_severity :debug, "Sleeping for #{interval} seconds"
           procline paused? ? "Paused" : "Waiting for #{queues.join(',')}"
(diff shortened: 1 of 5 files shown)


jeremywadsack:
This test failure is related to an update to Airbrake. I can reproduce it on `master` with the latest Airbrake. I've narrowed the issue down to a change in `airbrake` gem between 8.1.4 and 8.2.0 which changed the dependency of `airbrake-ruby` from 3.2.6 to 4.1.0. My guess is it has something to do with airbrake/airbrake-ruby#429 or airbrake/airbrake-ruby#443. I tried configuring Airbrake to ignore the environment for the rake tests, but even though it accepts the configuration it still reports as being "closed".

lib/resque/worker.rb
```diff
@@ -11,6 +12,7 @@ module Resque
   # It also ensures workers are always listening to signals from you,
   # their master, and can react accordingly.
   class Worker
+    include Observable
```
jeremy: How is Observable being used here? I'm not seeing changed/notify_observers.

jeremywadsack: Good catch. I'll get that fixed up early next week,

jeremywadsack:
@jeremy Removed that spurious `include` and rebased so the tests pass. Let me know if you want me to squash commits.

lib/resque/worker.rb
```diff
@@ -1,3 +1,4 @@
+require 'observer'
```
jeremy: Drop this require too?

jeremy:
Thanks! Looking good. One last observer require to nix. Please do squash 👍🏼

jeremywadsack:
Thanks @jeremy. Fixed that last issue and squashed. I also noticed that the test for "will call the queue empty hook only once" was not testing that it was called _just_ once, so I fixed that as well.

Let me know if you see anything else. Excited to have these hooks available!

alexcameron89:
Awesome work here @jeremywadsack, I'm excited to take advantage of these new hooks. 👏 👏

PR #1687, opened 2019-07-10
Escape id parameter for queues view

### LINKED ISSUE #1679: Vulnerable to Reflected Cross Site Scripting through pathname
`https://example.com/queues/`

image

As shown in the image above, the application is vulnerable to Reflected Cross Site Scripting through pathname of the site. With Non-Persistent cross-site scripting, malicious code is executed by the victim’s browser, and the payload is not stored anywhere; instead, it is returned as part of the response HTML that the server sends. The vulnerability is typically a result of incoming requests not being sufficiently validated, which allows for the manipulation of a web application’s function and the activation of malicious scripts. In order to exploit this vulnerability, an attacker needs to send the link to the victim. When the victim loads the link, the cross-site scripting payload will execute.

**Recommended fix:**
 HTML ESCAPE: Ruby on Rails has a built-in XSS protection mechanism which automatically HTMLescapes all the data being transferred from Rails to HTML. Rails uses ERB::Util#html_escape function toescape HTML entities. This particular function does the substitution based on the following hash (moreabout ERB::Util). ``` { '&' => '&amp;', '>' => '&gt;', '<' => '&lt;', '"' => '&quot;', "'" => '&#39;' } ``` Here is anexample: ``` <b>Hello <%= params[:name] %></b> ``` If the above is attacked with ?name=<script>alert(1)</script>, the output will be as follows: ``` <b> Hello &lt;script&gt;alert(1)&lt;/script&gt;gt; </b> `

brianvans:
Fixes #1679

### DIFF
--- a/lib/resque/server/views/queues.erb
+++ b/lib/resque/server/views/queues.erb
@@ -2,7 +2,7 @@
 
 <% if current_queue = params[:id] %>
 
-  <h1>Pending jobs on <span class='hl'><%= current_queue %></span></h1>
+  <h1>Pending jobs on <span class='hl'><%= h current_queue %></span></h1>
   <form method="POST" action="<%=u "/queues/#{current_queue}/remove" %>" class='remove-queue'>
     <input type='submit' name='' value='Remove Queue' onclick='return confirm("Are you absolutely sure? This cannot be undone.");' />
   </form>


lib/resque/server/views/queues.erb
```diff
@@ -2,7 +2,7 @@
 
 <% if current_queue = params[:id] %>
 
-  <h1>Pending jobs on <span class='hl'><%= current_queue %></span></h1>
+  <h1>Pending jobs on <span class='hl'><%= h current_queue %></span></h1>
```
iloveitaly, line 5: `h` is only available in rails and resque does not depend on rails. This should do the trick:

```suggestion
  <h1>Pending jobs on <span class='hl'><%= ERB::Util.html_escape current_queue %></span></h1>
```

@brianvans does this look good to you?

iloveitaly:
@brianvans what do you think of the suggested change? If you commit/accept, I can work on getting this merged.

lib/resque/server/views/queues.erb
brianvans: `h` is aliased as a helper over here:

https://github.com/brianvans/resque/blob/70f14b12d495ac0727cfe69fbfa2209fe2c42a31/lib/resque/server.rb#L28-L30

I noticed it was used in a few other views, so it made sense to use it here as well.

iloveitaly: Ah, nice! Didn't realize that.

iloveitaly (approved):

iloveitaly:
@chrisccerami Here's another one ready for merging!

iloveitaly:
@brianvans can you rebase on master so we can get the tests to pass?

brianvans:
> @brianvans can you rebase on master so we can get the tests to pass?

Done!

iloveitaly (approved):
@chrisccerami could you merge this one in?

G-Rath:
@brianvans @iloveitaly could we get a security advisory published for this?

G-Rath:
cc @corincerami just realised you're the one that did the merging

PR #1694, opened 2019-10-26
[ready] Improve logging instruction

tiendo1011:
Improve logging instruction so that users can just copy code and it'll work as theirs intended.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 .ruby-version
 coverage
 .rbx
-Gemfile.lock
 test/dump.rdb
 test/dump-cluster.rdb
 test/stdout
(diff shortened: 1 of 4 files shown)


tiendo1011:
@rafaelfranca Can you take a look at this?

iloveitaly:
@tiendo1011 what problem does this PR solve for you? Could you explain this further?

We don't want to recommend a non-monologger logger since it will cause logging issues in various corner cases (where signals are sent to the process).

iloveitaly:
Closing this out in favor of https://github.com/resque/resque/pull/1729 where we are making some improvements to the logging documentation.

PR #1699, opened 2019-12-20
Enable Rails logger to be displayed in STDOUT by default in the devel…

piecehealth:
…opment environment.

I saw a lot of developers feel confused for Resque log in the development environment. Just set it to STDOUT by default and make the consistent logger experience between `rails server` and `rails resque:work`.

Most of the code is referenced from https://github.com/rails/rails/blob/bf625f7fecabbcda22b388e088ad5c29016b2385/railties/lib/rails/commands/server/server_command.rb#L76-L86

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,7 +15,7 @@ Layout/Tab:
   Enabled: true
 
 # Blank lines should not have any spaces.
-Layout/TrailingBlankLines:
+Layout/TrailingEmptyLines:
   Enabled: true
 
 # No trailing whitespace.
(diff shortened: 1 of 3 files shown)


tsrivishnu:
Would like to see this merged. Is something stopping this from being merged?

iloveitaly:
@piecehealth @tsrivishnu I love the idea of documenting the logging configuration more, but I don't think it's a good idea to modify the core Rails logging configuration within Resque. This is something that should be done in the application configuration. 

Something that would be helpful if adding a section to [the logging documentation](https://github.com/resque/resque#logging) helping users redirect logs to the Rails logger in development.

Going to close this out, but feel free to submit another PR to improve documentation about logging with rails.

PR #1715, opened 2020-04-08
Add railtie

josh-m-sharpe:
This should eliminate some basic rails boilerplating.  It should also be non-breaking for any apps that currently override `resque:setup` in their app.


iloveitaly (approved):
Great change! Thanks for the contribution.

iloveitaly:
@jeremy Approved this change, I believe it's ready to merge.

PR #1732, opened 2020-10-30
Add button for deleting jobs from queue

h3nnn4n:
More than once I was required to either remove one pending job from Resque, or all jobs of a given class. For both cases I had to access the production server and run a custom script, which is not ideal. Based on that, this PR makes two changes:
1) Adds a dropdown with the list of all job classes in a queue and a button that removes all jobs of the selected class from the queue.
2) A button next to a pending job to delete that specific job

Here is a screenshot of how it looks:
Resque

Some notes:
* To properly support `ActiveJob` jobs I had to add some checks to fetch the class from the payload
* The button to remove a job will delete all jobs matching class and args. I could not find a better way to do this.

### DIFF
--- a/lib/resque/server.rb
+++ b/lib/resque/server.rb
@@ -185,7 +185,11 @@ module Resque
     end
 
     post "/queues/:queue/remove-jobs" do
-      Resque::Job.destroy(params[:queue], params[:job_class])
+      if params[:job_args].present?
+        Resque::Job.destroy(params[:queue], params[:job_class], *JSON.parse(params[:job_args]))
+      else
+        Resque::Job.destroy(params[:queue], params[:job_class])
+      end
       redirect url_path("/queues/#{params[:queue]}")
     end
 
--- a/lib/resque/server.rb
+++ b/lib/resque/server.rb
@@ -184,6 +184,11 @@ module Resque
       redirect u('queues')
     end
 
+    post "/queues/:queue/remove-jobs" do
+      Resque::Job.destroy(params[:queue], params[:job_class])
+      redirect url_path("/queues/#{params[:queue]}")
+    end
+
     get "/failed/?" do
       if Resque::Failure.url
         redirect Resque::Failure.url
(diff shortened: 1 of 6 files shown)


lib/resque/server/views/job_class.erb
```diff
@@ -1,6 +1,12 @@
+<form method="POST" action="<%=u "/queues/#{current_queue}/remove-jobs" %>" class="inline_form">
+  <input id="job_class" name="job_class" type="hidden" value=<%=job['class']%> />
+  <input id="job_args" name="job_args" type="hidden" value=<%=job['args'].to_json%> />
```
eduardoborba, line 3:
```suggestion
  <input id="job_class" name="job_class" type="hidden" value=<%= job['class'] %> />
  <input id="job_args" name="job_args" type="hidden" value=<%= job['args'].to_json %> />
```

eduardoborba (approved):

sandipsubedi (approved):

matthewhively:
It would be nice to have this feature.

lib/resque/server.rb
```diff
@@ -73,6 +184,15 @@ def show_for_polling(page)
       redirect u('queues')
     end
 
+    post "/queues/:queue/remove-jobs" do
+      if params[:job_args].present?
```
PatrickTulskie, line 188: This requires ActiveSupport, which we may not always have available.

PatrickTulskie (changes requested):
@h3nnn4n Hey I know this has been sitting for the better part of 6 years. I'm happy to get it merged if you're still interested. It could use a rebase and an update. Also, if you can remove the HISTORY.md change so that there aren't merge conflicts.

PatrickTulskie:
I'm going to close this out as superseded by #1969 

We added a delete button for individual jobs in there, but did not add the "Remove Jobs" button at the top. Simple reason is that mass job deletion is brutal in resque as it sits. We also opened up ticket #1971 but I want to hold that until some variation of #1788 ships which may also require #1970 depending on what we settle on for architecture.

In any case, things are moving. The per-job delete button will go out in 3.1 and then we'll do some further optimizations to potentially land in 3.2. Thanks for the idea and your contribution!

PR #1733, opened 2020-11-07
Ensure PIDFILE has pids of all running workers

### LINKED ISSUE #1721: PIDFILE contains only a single PID after starting multiple workers
When starting multiple workers using rake task `resque:workers` and the `PIDFILE` env variable, only a single pid is available in the specified file, although multiple workers were started.

z0marlin:
The workers now append to the PIDFILE instead of overwriting it.
Adding file locking to handle race conditions when multiple
workers are started using resque:workers.

To clear any older data in the specified PIDFILE, it is now truncated
once when the rake tasks resque:work and resque:workers are called.
Created a proxy task resque:run to achive this.

Fixes #1721

### DIFF
--- a/lib/resque/tasks.rb
+++ b/lib/resque/tasks.rb
@@ -5,8 +5,8 @@
 namespace :resque do
   task :setup
 
-  desc "Start a Resque worker"
-  task :work => [ :preload, :setup ] do
+  desc "Run a Resque worker"
+  task :run => [ :preload, :setup ] do
     require 'resque'
 
     begin
@@ -20,6 +20,15 @@ namespace :resque do
     worker.work(ENV['INTERVAL'] || 5) # interval, will block
   end
 
+  desc "Start a Resque worker"
+  task :work do
+    if ENV['PIDFILE'] && File.file?(ENV['PIDFILE'])
+      File.truncate(ENV['PIDFILE'], 0)
+    end
+    Rake::Task['resque:run'].reenable
+    Rake::Task['resque:run'].invoke
+  end
+
   desc "Start multiple Resque workers. Should only be used in dev mode."
   task :workers do
     threads = []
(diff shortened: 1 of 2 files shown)


Z0Marlin:
ping

lib/resque/tasks.rb
```diff
@@ -20,6 +20,15 @@
...
+  desc "Start a Resque worker"
+  task :work do
+    if ENV['PIDFILE'] && File.file?(ENV['PIDFILE'])
+      File.truncate(ENV['PIDFILE'], 0)
+    end
+    Rake::Task['resque:run'].reenable
+    Rake::Task['resque:run'].invoke
+  end
+
   desc "Start multiple Resque workers. Should only be used in dev mode."
```
iloveitaly, line 32: Do you understand the use case for this? Do you use this in production? Resque is designed around forking, so it seems strange that we support this. Also, shelling out to a rake task feels a bit dangerous to me for production-quality deployments.

iloveitaly (changes requested):
Could you help me understand the use case here more fully? I don't see why we have this task in the first place.

To get this merged:

* Could you respond to the message about why we aren't building the PIDFILE logic into `run`?
* Could you add a changelog entry?
* Rebase on master so CI runs

Thanks for the contribution!

z0marlin:
>Could you help me understand the use case here more fully? I don't see why we have this task in the first place.

Hey @iloveitaly! The issue this change tries to fix is #1721. We have a requirement to automate starting and stopping of resque workers. When starting multiple workers in the background, to stop each of the workers, we need the PIDs of the workers and so it would be great to have all the PIDs in the PIDFILE.

This is quite an old PR. I'll need some time to look at the changes I made and then I can address the comments. 

Thanks for reviewing the PR 😄

iloveitaly:
> This is quite an old PR. I'll need some time to look at the changes I made and then I can address the comments.

@z0marlin sound great!

PatrickTulskie:
Thanks for this, and apologies that it went quiet for so long.

That said, I'm going to close it, mainly over the PIDFILE semantics change: going from a single pid to newline-delimited pids will break anything currently reading that file expecting one value. It also restructures the task graph (resque:work becomes a wrapper around a new resque:run), which downstream Rakefiles may depend on. That's a wider compatibility surface than I want to take on without a major release or at least some sort of migration process for people.

The use case is reasonable, though. If you're still interested, I'd happily look at a fresh PR against current master that either writes a separate file for the multi-worker case and/or puts the new behavior behind an opt-in, so existing PIDFILE consumers are left alone.

In any case, I think for most multi-worker situations these days, people just use https://github.com/resque/resque-pool

PR #1734, opened 2020-11-12
COUNT=2 should say "two" instead of "five"

rferreira:
I'm 70% sure the documentation is wrong


sandipsubedi (approved):

iloveitaly (approved):
Great catch, thanks for the fix!

PR #1735, opened 2020-12-11
Update README with Rails example

karpiu:
I know the Rails case is in the wiki, but it took me awhile to figure this out, and in the meantime I was stuck. I even posted a question on stackoverflow:

https://stackoverflow.com/questions/65224796/resque-not-failing-an-obviously-buggy-job

And one of the users suggested I request this change, as the Rails is pretty popular case for the Resque gem.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -137,7 +137,7 @@ access to the Redis server.
 
 #### Rails example
 
-If you installed the Resque as a Rails gem, the above example will look like this:
+If you are using ActiveJob here's how your job definition will look:
 
 ``` ruby
 class ArchiveJob < ApplicationJob
--- a/README.markdown
+++ b/README.markdown
@@ -135,6 +135,34 @@ Workers can be given multiple queues (a "queue list") and run on
 multiple machines. In fact they can be run anywhere with network
 access to the Redis server.
 
+#### Rails example
+
+If you installed the Resque as a Rails gem, the above example will look like this:
+
+``` ruby
+class ArchiveJob < ApplicationJob
+  queue_as :file_serve
+
+  def perform(repo_id, branch = 'master')
+    repo = Repository.find(repo_id)
+    repo.create_archive(branch)
+  end
+end
+```
+
+``` ruby
+class Repository
+  def async_create_archive(branch)
+    ArchiveJob.perform_later(self.id, branch)
+  end
+end
+```
+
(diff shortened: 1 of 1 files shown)


iloveitaly:
@karpiu this is great! I think it would fit better as the last section within the "Jobs" header as opposed to right after the intro paragraph:

https://github.com/resque/resque#jobs

Could you move this around and ping me again on the PR?

README.markdown
```diff
@@ -135,6 +135,34 @@ Workers can be given multiple queues (a "queue list") and run on
 multiple machines. In fact they can be run anywhere with network
 access to the Redis server.
 
+#### Rails example
+
+If you installed the Resque as a Rails gem, the above example will look like this:
```
iloveitaly, line 140: I think this should worded differently: resque is always going to be installed as a gem, and there's no such thing as a 'rails gem'. Maybe something like "If you are using ActiveJob here's how your job definition will look:"

iloveitaly (changes requested):
@karpiu this is great! I think it would fit better as the last section within the "Jobs" header as opposed to right after the intro paragraph:

https://github.com/resque/resque#jobs

Could you move this around and ping me again on the PR?

karpiu:
Addressed comments @iloveitaly

iloveitaly (approved):
Great! Thank you.

README.markdown
```diff
@@ -443,6 +443,35 @@ Keep this in mind when writing your jobs: you may want to throw
...
+``` ruby
+class Repository
+  def async_create_archive(branch)
+    ArchiveJob.perform_later(self.id, branch)
+  end
+end
+```
+
+It is important to run `ArchiveJob.perform_later(self.id, branch)` rather than `Resque.enqueue(Archive, self.id, branch)`.
+Otherwise the Resque will process the job without actually doing anything.
```
chrisccerami, line 470: I would remove the word `the` from between `Otherwise` and `Resque` to improve the grammar a bit.

README.markdown
```diff
@@ -443,6 +443,35 @@ Keep this in mind when writing your jobs: you may want to throw
...
+Even if you put obviously buggy line like `0/0` in the `perform` method,
```
chrisccerami, line 471: This is missing an article, like `Even if you put an obviously buggy line`.

karpiu:
@chrisccerami done

chrisccerami:
Thanks a lot @karpiu

PR #1736, opened 2020-12-18
upgrade jquery version to 3.5.1 to address vulnerabilities

hfadsk:
We currently use jquery version 1.12.4 which has the following vulnerabilities: CVE-2015-9251, CVE-2019-11358, CVE-2020-11022, CVE-2020-11023.

This PR upgrades the jQuery version to 3.5.1 to address those vulnerabilities.


iloveitaly:
@hfadsk thanks for the contribution! I'm not too familiar with the server code in Resque, but I would imagine there would be some breaking changes between `1.x.x` and `3.5.x`. The test suite does not cover any JS code from what I can tell.

Were you able to test this branch on a production system?

iloveitaly:
@hfadsk checking in here to see if you'll get a chance to work on this PR a bit more.

peret:
Hi @iloveitaly, we would love to see this PR merged. I will check it out and try to test it with our system. Is there anything else that needs to be done before this can be released?

peret:
I tested these changes in our environment and haven't found any issues so far. I made sure that timestamps are still converted to relative times, retrying and removing failed jobs still works, and all other features were I can see jQuery being used are also fine 👍

iloveitaly:
@peret That's great to hear. It does look our jQuery code is safe to upgrade after some quick inspection.

Couple things we need to do to get this merged:

* Upgrade to jquery 3.6.0
* Rebase on master
* Add entry to changelog

Do you have time to take a look?

peret:
@iloveitaly Great! Since I'm not the original author of this PR, I can't update it, but I will just create a new PR and reference it here.

peret:
It's #1763 ✅

PR #1740, opened 2021-03-05
live poller shouldn't restart itself until it succeeds or fails

jsharpe-ctm:


iloveitaly:
Thanks @jsharpe-ctm! Looks great.

I imagine the use case here is if the server is overloaded and taking longer to respond than the timeout interval, the server gets overloaded with many of the same requests?

iloveitaly (approved):

jsharpe-ctm:
@iloveitaly precisely - thanks!

PR #1747, opened 2021-04-26
Fix error class comment

matthewhively:
Just a comment


iloveitaly (approved):
Good catch! Thanks for the fix.

PR #1748, opened 2021-06-09
Fix failing build due to rubocop change

iloveitaly:
* Bump all tested ruby versions to last minor version
* Sets rubocop execution version to `2.7.3`

@rafaelfranca could you take a look at this one and merge? This is causing all PR builds to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ before_install:
 script: bundle exec rake
 
 rvm:
-  - 2.4.9
-  - 2.5.7
-  - 2.6.5
-  - 2.7.0
+  - 2.4.10
+  - 2.5.9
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,16 +8,16 @@ before_install:
 script: bundle exec rake
 
 rvm:
-  - 2.4.9
-  - 2.5.7
-  - 2.6.5
-  - 2.7.0
+  - 2.4.10
```
byroot, line 11: I'm oretty sure you can set `2.4` to automatically get the last available version.

rafaelfranca:
I'm going to merge this but we should consider moving to GitHub Actions.

PR #1752, opened 2021-06-28
Setup GitHub Actions

mtsmfm:
Feel free to close this PR if you're migrating to https://travis-ci.com.

-----

I've noticed travis-ci.org was ceased two weeks ago.
image

As a result, currently CI against new PRs like https://github.com/resque/resque/pull/1751 doesn't work.

I confirmed this setting works well:
https://github.com/mtsmfm/resque/pull/1


.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+          - 3.0.1
+          - head
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "${{ matrix.ruby-version }}"
+      - run: sudo apt-get install redis-server
+      - run: bundle install
+      - run: bundle exec rubocop
```
iloveitaly, line 27: Could we run rubocop only on a single build?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
```
iloveitaly, line 5: I think we want to run GH actions on every PR and any branch. No reason not to as far as I'm aware.

```suggestion
on: [push, pull_request]
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby-version:
+          - 2.4.10
+          - 2.5.9
+          - 2.6.7
+          - 2.7.3
+          - 3.0.1
```
iloveitaly, line 18: I think it's better here to drop the x.x.y and just specify x.x so the latest version is always used.

iloveitaly (changes requested):
This is great! Thank you for your work here.

Left a couple of comments. Could you also remove the travis yaml and remove the travis badge from the readme?

.github/workflows/ci.yml
mtsmfm: If we specify both and resque commiters creates new branch on resque/resque repo, both push and pull_request CI run.
Is it ok?

image

README.markdown
```diff
@@ -2,7 +2,7 @@ Resque
 ======
 
 [![Gem Version](https://badge.fury.io/rb/resque.svg)](https://rubygems.org/gems/resque)
-[![Build Status](https://travis-ci.org/resque/resque.svg)](https://travis-ci.org/resque/resque)
```
mtsmfm, line 5: It'll be like:

[Build Status](https://github.com/mtsmfm/resque/actions/workflows/ci.yml)

mtsmfm:
@iloveitaly Thank you for reviewing.
I fixed ci.yml, removed travis.yml and migrated status badge to GitHub Actions' one.

But I'm not sure if this `on` setting is what you intended.
https://github.com/resque/resque/pull/1752#discussion_r664163306

Can I ask you to review again?

.github/workflows/ci.yml
iloveitaly: Ah, wasn't aware that the build is run twice in that case. I think it's helpful to developers to run CI against their branch before opening a PR, but probably makes sense to use the default here instead. Thanks for calling this out—feel free to change it back.

README.markdown
iloveitaly: That's great! Thanks for adding the GH badge.

iloveitaly (changes requested):
You were right about `on`, could you revert that back to the previous config? Otherwise, this looks great! Thank you.

mtsmfm:
@iloveitaly Reverted. Can you review again? Thanks!

iloveitaly (approved):

PR #1762, opened 2021-08-09
Make server safe to use with Content Security Policies that don't allow 'unsafe-inline'.

### LINKED ISSUE #1756: Content Security Policy triggers on Resque frontend
We're testing a new nonce-based Content Security Policy for our Rails app. The CSP does not contain 'unsafe-inline', which seems to break some inline script in the Resque frontend of our app. In Firefox, I get the error `Content Security Policy: The page's settings observed the loading of a resource at inline ("script-src"). A CSP report is being sent. Source: t.`. The CSP is in "report-only" mode for now.

Is there any way for me to work around this? I already tried finding out whether there's a way for me to disable/change the CSP for the resque-related routes, but so far didn't find a way to e.g. trigger a before_action for the routes of the Engine.
Or is this maybe something that can be fixed in the gem? I've seen another gem offering a config option to override the CSP, although I'm not sure that's the best option. I think the ideal solution would be to not use any inline-scripts in the views at all.

peret:
Replace inline onclick handlers with a jQuery click handler registered
on a new class 'confirmSubmission'.

Fixes #1756.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -11,6 +11,7 @@
 ### Fixed
 * Prune workers which haven't been registered but have set a heartbeat
 * `Resque::Failure::Multiple.remove` did not pass on the queue parameter
+* Replace onclick handlers in server code to support Content Security Policies that don't allow 'unsafe-inline'
 
 ## 2.0.0 (2018-11-06)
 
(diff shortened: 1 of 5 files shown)


iloveitaly:
@peret could you add an entry to the changelog about this change? It looks great!

peret:
Done 👍

PR #1768, opened 2021-09-23
Fix 'uninitialized constant Resque::Failure::RedisMultiQueue' on `/overview`

sharang-d:
When the `Resque::Failure::RedisMultiQueue` backend is not used the resque-web page throws an exception.
I think this started happening with https://github.com/resque/resque/pull/1638.

Steps to reproduce:
```
cd resque/examples/demo
rackup config.ru
# Open http://localhost:9292/resque
```

Before:
overview 2021-09-24 02-07-21

After:
Resque 2021-09-24 02-08-05

I'd like to write a spec to verify the fix but I can't find a way of **not** including the failure class in the resque-web test file. Tips if any would be appreciated :)

I found this problem when I began looking into why specs were failing in https://github.com/resque/resque-scheduler/pull/671#issuecomment-924974183.

### DIFF
--- a/lib/resque/server/helpers.rb
+++ b/lib/resque/server/helpers.rb
@@ -11,7 +11,7 @@ Resque::Server.helpers do
     return @multiple_failed_queues if defined?(@multiple_failed_queues)
 
     @multiple_failed_queues = Resque::Failure.queues.size > 1 ||
-      Resque::Failure.backend == Resque::Failure::RedisMultiQueue
+      (defined?(Resque::Failure::RedisMultiQueue) && Resque::Failure.backend == Resque::Failure::RedisMultiQueue)
   end
 
   def failed_size


sharang-d:
cc @tsevan @iloveitaly as you may have more context around this.

tsevan:
Thanks @sharang-d ! That indeed looks like a sufficient hotfix. Alternatively we could require the file with the class definition somewhere in the helpers section: `require 'resque/failure/redis_multi_queue'`. No idea though how much safe it is.

Let's wait for the repo maintainers to review the change.

sharang-d:
I didn't do that because it didn't feel right to bring in the class just to compare it with something (and not "use" it).

iloveitaly:
This approach looks fine to me! Thanks for the fix.

Issue #1771, opened 2021-09-28
Update/remove homepage or move docs over?


fatso83:
Hi, this repo lists resque.github.io as its homepage, but there is nothing there. Might as well either remove the link or perhaps use the gh-pages feature to move the docs over, as I was looking there for notes on installation, etc :)

iloveitaly:
@fatso83 great idea! Could you submit a PR to move things over there?

Also, would be great to move some of the more in-depth documentation in the wiki over to the pages setup.

iloveitaly:
More notes about what we could do here: https://github.com/resque/resque/issues/1759#issuecomment-966443768

PR #1778, opened 2021-12-03
README suggestions

varyonic:
In response to #1771 

Also, should HOOKS and PLUGINS move to the docsite?


iloveitaly:
Nice round of improvements!

Yes, if we could move the other docs to the resque site—along with some of the content in the wiki—that would be great.

PR #1781, opened 2021-12-21
Add rel attribute in anchor link to mitigate reverse tabnabbing

saran1856:
We use RubyGem-resque as a dependency for one of our services. 

Our company's InfoSec org has called out a reverse tabnabbing attack in "job_class.erb" and in order to use it, we have to add "rel" attribute to value "noopener noreferrer".

Here's a demo and example explaining how this change will fix the issue: https://mathiasbynens.github.io/rel-noopener/


iloveitaly:
Thanks for the contribution!

PR #1784, opened 2021-12-28
Update peek documentation to better represent type variance based on count

jharrilim:
# Update peek documentation to better represent type variance based on count

Just a small doc change to mention that the item itself is returned when count is 1, as opposed to an array of 1 item


iloveitaly:
Thank you!

PR #1790, opened 2022-02-21
Escape HTML from the params to avoid XSS

madslundholmdk:
After a penetration test, there were multiple Reflected XSS vulnerabilities found in the the web interface.

The following paths has been found vulnerable:
- `/failed/?class=<script>alert(document.cookie)</script>`
- `/queues/>`

I have look through the other views, and found other places where HTML wasn't escaped from the params.


varyonic (approved):

iloveitaly:
Great change, thank you!

Issue #1664, opened 2019-01-10
Default `exit!` produces unexpected behavior


jweir:
There have been several previous issues around the use of `exit!` https://github.com/resque/resque/issues/1598 https://github.com/resque/resque/issues/1412 https://github.com/resque/resque/issues/1483 https://github.com/resque/resque/issues/1167 https://github.com/resque/resque/issues/1646

Back with https://github.com/resque/resque/pull/307#issuecomment-1271733 Resque tried to change `exit!` to `exit` (ps @jeremy thank you for tracking down the change!)

Looking at the original code here, there is an `ensure` block on line 145 which would get called via `exit` – this was not desired.

https://github.com/resque/resque/blob/f3bbf9916e515a7817d6d592c4549ffde62774b2/lib/resque/worker.rb#L130-L146

But the current code is much simpler.  The forked process calls `perform` and is done.  There no longer is an `ensure` block or any additional code to run. 

https://github.com/resque/resque/blob/1d85f7c705ee8a609531c29035f8cf7cc6d67b19/lib/resque/worker.rb#L895-L899

So, switching from `exit!` to `exit` should be safe – except for applications which have `at_exit` blocks that currently don't get called.

Seeing that there have been several issues around this, and we spent a good deal of time tracking this down, I think it should be fixed.

I see two options.  

A) Change the default behavior to `exit`, remove the `RUN_AT_EXIT_HOOKS` env option, and decide if `exit!` should be an option.

B) Keep the behavior, and document the work around up front.

I'm in favor of `A` and would be happy to put together a PR.

stuarthannig:
+1

stayhero:
Yes, 👍 for A.

jfragoulis:
How is this coming along?

st0012:
I think Resque should switch to `exit` to allow `at_exit` callbacks. The current behavior breaks libraries that rely on the `at_exit` to perform cleanup work (like [sentry-ruby](https://github.com/getsentry/sentry-ruby)) in an implicit way. 

And because related issues would be hard to debug, I think the readme should also document about this behavior. I'll open a PR for that shortly.

PR #1795, opened 2022-03-14
Mention `at_exit` and `RUN_AT_EXIT_HOOKS` env var in the readme

st0012:
As reported in https://github.com/resque/resque/issues/1664, the default behavior of using `exit!` can cause issues that are hard to debug. So even if it's not a subject to change, this gotcha and the related option should be mentioned in an obvious place.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -687,6 +687,12 @@ complicated.
 
 Workers instead handle their own state.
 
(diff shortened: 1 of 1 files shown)


iloveitaly:
Great addition, thank you!

PR #1799, opened 2022-04-22
Handle when Rails is partially loaded as a Gem

TSMMark:
and not an application

```
NoMethodError: undefined method `application' for Rails:Module
    /bundle/ruby/3.0.0/gems/resque-2.0.0/lib/resque/tasks.rb:43:in `block (2 levels) in <main>'
```

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,8 @@
 
 ### Fixed
 
+* Fix NoMethodError: undefined method 'application' for Rails:Module when Rails module is defined but not a full Rails app (#1799)
+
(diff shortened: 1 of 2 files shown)


iloveitaly:
@TSMMark in what scenarios could this happen?

In this case, wouldn't want an error to be thrown so the dev could fix the load order?

iloveitaly:
Ah, that makes sense! Could add a changelog entry for this to the PR? I'll merge it in!

TSMMark:
@iloveitaly done!

PR #1802, opened 2022-06-07
Add Ruby syntax highlighting to rest of markdown docs

dsinn:
For consistency with README.markdown, and because it's easier to read IMO.


iloveitaly:
Great change, thanks!

PR #1803, opened 2022-06-10
Added support for pausing the queue by setting a Redis key

pietervogelaar:
At the moment it is only possible to pause a specific worker with a signal. It would be nice to pause all queue workers, so that they finish their current job, but don't pickup a new job. This pull request makes it possible to pause all queue workers by setting a Redis key:
```
Resque.redis.set('paused', 'true')
```
This way a deployment of new workers (from a deployment pipeline) can easily be done without aborting running jobs.
When the deployment or other maintenance is finished the queue workers can be unpaused by deleting the Redis key:
```
Resque.redis.del('paused')
```

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -579,7 +579,7 @@ module Resque
 
     # are we paused?
     def paused?
-      @paused
+      @paused || redis.get('paused') == 'true'
     end
 
     # Stop processing jobs after the current one has completed (if we're
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -579,7 +579,7 @@ module Resque
 
     # are we paused?
     def paused?
-      @paused || redis.get('paused') == 'true'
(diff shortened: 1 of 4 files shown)


lib/resque/worker.rb
```diff
@@ -579,7 +579,7 @@ def wait_for_child_exit(timeout)
 
     # are we paused?
     def paused?
-      @paused
+      @paused || redis.get('paused') == 'true'
```
iloveitaly, line 582: Can we strip + downcase the value here (so `TRUE` works and not just `true`).

iloveitaly (changes requested):
I think this is a good idea since we already support it via signals. Some changes I'd like to see before merging:

1. We should use a more descriptive key `pause-all-workers` or something along those lines
2. Can you add a changelog entry?
3. Can you add docs to the readme?
4. Can we add a test for this?

pietervogelaar:
@iloveitaly Thank you! I committed the requested changes.

PR #1814, opened 2022-08-07
Matrix test redis version

iloveitaly:


Gemfile
```diff
@@ -1,9 +1,14 @@
 source "https://rubygems.org"
-
 gemspec
 
-gem "json"
+case redis_version = ENV.fetch('REDIS_VERSION', 'latest')
+when 'latest'
+  gem 'redis', '~> 4.7'
+else
+  gem 'redis', "~> #{redis_version}"
```
nevans, line 8: I think you'll need to change this, to avoid loading a higher than desired version.  E.g:
```suggestion
  gem 'redis', "~> #{redis_version}.0"
```

If you look at the CI logs, you can see that it isn't installing the requested major.minor version, but the latest (installable) major version.

That's in addition to adding some matrix exclusions for old rubies...

.github/workflows/ci.yml
```diff
@@ -12,15 +12,27 @@ jobs:
...
         ruby-version:
           - 2.3
           - 2.4
           - 2.5
           - 2.6
           - 2.7
-          - "3.0"
+          - 3.0
           - 3.1
           - head
```
nevans, line 32: The latest redis gems don't support ruby 2.3.  Need to add exclusions to the test matrix... or drop support for ruby 2.3. 😉

```suggestion
          - head
        exclude:
          - { ruby-version: 2.3, redis-version: 4.5 }
          - { ruby-version: 2.3, redis-version: 4.6 }
          - { ruby-version: 2.3, redis-version: latest }
```

.github/workflows/ci.yml
```diff
@@ -12,15 +12,27 @@ jobs:
...
+          - 4.6
+          - latest
         ruby-version:
           - 2.3
           - 2.4
           - 2.5
           - 2.6
           - 2.7
-          - "3.0"
+          - 3.0
```
nevans, line 30: Has https://github.com/actions/runner/issues/849 been fixed?

At least in the past, it was better to keep the quotes.  Otherwise, it's treated as a number instead of a version string.  And `3.0` can be interpreted as `3`, which can mean "the latest 3.x version".  If the redis gem ever has a "4.10" version that could be interpreted as "4.1", etc.  If consistency is what's wanted, you can add quotes to *all* of the versions.

```suggestion
          - "3.0"
```

nevans (commented):
looks good!  I have some suggestions, for your consideration. 🙂

.github/workflows/ci.yml
iloveitaly: You are right, figured this was fixed, but it isn't: https://github.com/resque/resque/runs/7859272121?check_suite_focus=true

iloveitaly:
Thanks for the suggestions @nevans!

PR #1816, opened 2022-08-09
Fix ordering of web-runner code

essjayhch:
Corrects the ordering of Namespace and config blocks, otherwise the namespace setting is lost if you pass redis string.


iloveitaly:
@essjayhch Thank you! Can you create another PR with a changelog entry?

PR #1818, opened 2022-08-20
Added flexible eager load configuration for Resque when used in Rails application

georgiybykov:
Hi!

We faced a problem when upgraded from Resque `1.27.4` to `> 2.0.0` version (using Resque gem as a dependency in Rails 6.0.4 application with Ruby 2.7.5) and lost many important events silently. The eager load for Resque task did not work was the reason.

Our production configuration for `eager_load` is set to `true`:
```ruby
# config/environments/production.rb

Rails.application.configure do
  ...
  config.eager_load = true
  ...
end
```

Changes documentation between versions says the following:

image

image

However, the `eager_load` will never be `true`. Since `Resque` starts from Rake task, the `eager_load` option configuration is set to `false` permanently by default starting from [this commit](https://github.com/rails/rails/pull/11389/files#diff-84292fcaae327f1a70e7c3b1ebb55b193dea016c1118a8566930bc5cd2a51ba5) at Rails.

Only if your Rails application is `6.1.0` version or above, you are able to configure eager loading for Rake tasks environment by using `rake_eager_load` option that was added in [this commit](https://github.com/rails/rails/pull/28209/files#diff-84292fcaae327f1a70e7c3b1ebb55b193dea016c1118a8566930bc5cd2a51ba5) to provide the ability to override the default `false` value **for every rake task of the environment**.

### Based on the above, I propose to add flexible configuration for `eager_load`:

To configure `eager_load` option using `Resque` interface you are able to add in `config/initializers/resque.rb`
the following:
```ruby
Resque.rails_eager_load_enabled = true # one configuration for any of your application environments
```

Or if you would like to configure the `eager_load` option depends on your Rails application environment, for example:
```ruby
Resque.rails_eager_load_configure do |env|
  env.development = false
  env.staging = true
  env.production = true
end
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .ruby-version
+.ruby-gemset
 coverage
 .rbx
 Gemfile.lock
@@ -7,7 +8,5 @@ test/dump-cluster.rdb
 test/stdout
 *.swp
 pkg
-.ruby-version
-.ruby-gemset
 *.sw?
 vendor/bundle
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -72,6 +72,6 @@ helpful.
 Help
 ----
 
-If you have a question or want to discuss something, the Resque mailng list
+If you have a question or want to discuss something, the Resque mailing list
 might just be the place. [resque@librelist.com](mailto:resque@librelist.com)
 is the address you want, send an email there and it'll take care of you.
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,17 @@
+## Unreleased
+
+### Fixed
+
+* Fix - Run `eager_load!` only if `Rails.application.config.eager_load` is true - from 2.0.0 version release
+  by adding flexible configuration. (#1818)
+
+### Added
+
+* `Resque::Railtie::EagerLoad` and `Resque::Railtie::EagerLoad::Configuration` classes for flexible eager load
+  configuration for Resque when used in a Rails application (#1818)
+
+*
+
 ## 2.5.0
 
 ### Fixed
@@ -82,7 +96,7 @@
 
 ### Fixed
 
-* live poller shouldn't restart itself until it successds or fails. #1740
+* live poller shouldn't restart itself until it succeeds or fails. #1740
 * Fix parsing worker_id when queue name includes colon. #1691
 * Prune workers which haven't been registered but have set a heartbeat. #1751
 * `Resque::Failure::Multiple.remove` did not pass on the queue parameter
--- a/README.markdown
+++ b/README.markdown
@@ -144,6 +144,42 @@ end
 We don't want the `git_timeout` as high as 10 minutes in our web app,
 but in the Resque workers it's fine.
 
+#### Eager load configuration for Rails (`Resque::Railtie::EagerLoad`)
+
+Since `Resque` starts from the Rake task, the configuration of the `eager_load` option is permanently set to `false` by default
+from [this commit](https://github.com/rails/rails/pull/11389/files#diff-84292fcaae327f1a70e7c3b1ebb55b193dea016c1118a8566930bc5cd2a51ba5)
+at Rails. If your Rails application is `6.1.0` version or higher, you can configure eager loading for
+the Rake tasks environment using the `rake_eager_load` option that was added in [this commit](https://github.com/rails/rails/pull/28209/files#diff-84292fcaae327f1a70e7c3b1ebb55b193dea016c1118a8566930bc5cd2a51ba5)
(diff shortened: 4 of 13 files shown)


iloveitaly:
@georgiybykov Is there a reason you don't do something like:

```
namespace :resque do
  task :setup => :environment do

  end
end
```

Pretty sure this would force rails to load before the resque app?

georgiybykov:
The file `lib/tasks/resque.rake` in our Rails application contains the following:
```ruby
# frozen_string_literal: true

require 'resque/tasks'

namespace :resque do
  task :setup => :environment

  ...
end
```

The code above is not a solution (does not force preload all classes for the application).

@iloveitaly As an alternative, do you suggest to override the `setup` task to force preload all classes in memory when booting the app per below? 
```ruby
# frozen_string_literal: true

require 'resque/tasks'

namespace :resque do
  task :setup => :environment

  task :preload => :setup do
    ActiveSupport.run_load_hooks(:before_eager_load, Rails.application)
    Rails.application.config.eager_load_namespaces.each(&:eager_load!)
  end
end
```

[Since Rake tasks ignore the Rails application `eager_load` configuration setting](https://github.com/rails/rails/pull/11389):

image

`eager_load` leads to `false` value here:

image

Only if you use Rails `>= 6.1.0`, you can impact on this setting via `Rails.application.config.rake_eager_load=`. However, it will be applied for all rake tasks of you application for current environment.

That is why in this pull request I suggest to add flexible eager load configuration for Resque when used in Rails application without actions to override the code in Rake tasks in order to setup correctly.

iloveitaly:
@georgiybykov your explanation makes sense. Tricky problem! I need to set aside some more time to review your PR. Will hopefully get back to you soon.

georgiybykov:
Hi @iloveitaly!

Thank you again! I would be very grateful for your feedback and I hope you get a chance to review the code soon! Please let me know if I can provide anything else in the meantime.

fxn:
I suspect there's a bit of confusion here:

* In Rake tasks, `config.eager_load` has been `false` by default for a long time.
* In older versions, this was overridden by hand, see [`4.0`](https://github.com/rails/rails/blob/4-0-stable/railties/lib/rails/application.rb#L258) for example.
* In newer version, as it is said above, users can set `config.rake_eager_load`, whose value will be assigned to `config.eager_load` before initialization, and it is `false` by default.

However:

* If `config.eager_load` is `true`, Rails already does what this Rake task is doing.
* Since idempotence is not a requirement, you'll indeed run stuff twice!

I believe this task should be deleted, though I might be missing something. Let Rails eager load, and let users control this via Rails. I opened https://github.com/resque/resque/pull/1867 with this proposal.

georgiybykov:
@fxn Thank you very much for your comment!

My opinion and the main points of the current pull request are:

- Prior to the 2.0.0 release change, users did not need to override the task manually to eagerly load the namespaces, as the `setup` task implemented this behaviour.
- If we simply delete the code for this task, users will have to override the task in their projects (if they are using Rails with a version below 6.1.0), and they should know about this "feature" by default, but the documentation says nothing about it.
- In my opinion, even if the application is using one of the latest versions of Rails, not all of them need to enable `rake_eager_load` which would affect every Rake task just because it uses Resque. So this request is about:
   - removing the confusion from the 2.0.0 release that points to `eager_load` configuration changes;
   - suggesting to add a more flexible eager load configuration using the Rasque interface.

___

> I believe this task should be deleted, though I might be missing something. Let Rails eager load, and let users control this via Rails. I opened https://github.com/resque/resque/pull/1867 with this proposal.

Based on the above, you are correct, however the users with Rails version `< 6.1.0` should always override the task:
```ruby
task :preload => :setup do
  ActiveSupport.run_load_hooks(:before_eager_load, Rails.application)
  Rails.application.config.eager_load_namespaces.each(&:eager_load!)
end
```
because they cannot control this via the Rails configuration.

PatrickTulskie:
Thanks for this, and sorry it sat so long. Took a minute to digest everything in the PR.

The problem you hit is real, but I think the cause is narrower than this PR assumes. Under a Rake task, Rails sets config.eager_load = config.rake_eager_load, and rake_eager_load defaults to false. So by the time resque's :preload task runs, config.eager_load is already false and the task no-ops — which is exactly the silent behavior you were seeing.

For most people the fix is one line in config/application.rb:

```ruby
config.rake_eager_load = true
```

With that set, Rails eager loads during boot and resque's :preload task is redundant, which is what #1867 removes.

Given that, I'd rather not add a separate eager-load configuration surface to resque itself. Closing in favor of #1867 plus a README note about rake_eager_load. I do appreciate the writeup and the repro bc it made the actual cause a lot easier to pin down.

If you think I'm off base here, just comment and lemme know.

PR #1822, opened 2022-08-29
Remove thread_safe arg in Redis instantiations

sl0thentr0py:
The new redis gem 5.0.x checks for valid args and throws on this call.

Fixes #1821

PR #1823, opened 2022-08-29
Pin redis version explicitly to <5

### LINKED ISSUE #1821: Unknown keyword: thread_safe with redis 5.0.1
The `redis` gem just had a new breaking release which seems to cause this to fail.

```
      Failure/Error: Resque.redis.del "queue:default"

      ArgumentError:
        unknown keyword: thread_safe
```

Maybe we need to pin the `redis` dependency here to < 5
https://github.com/resque/redis-namespace/blob/master/redis-namespace.gemspec#L30

sl0thentr0py:
As an alternative to https://github.com/resque/resque/pull/1822, just pin the redis version to avoid the new breaking changes.
Fixes #1821

### DIFF
--- a/resque.gemspec
+++ b/resque.gemspec
@@ -23,6 +23,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.3.0"
 
   s.add_dependency "redis-namespace", "~> 1.6"
+  s.add_dependency "redis", ">= 4", "<5"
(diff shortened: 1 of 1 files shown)


nevans:
IMO this isn't an alternative to #1822.  They should both be merged and released.  There are additional issues beyond what #1822 fixes, but it's at least a first step.  And, even if those other issues are fixed and released today, I like the documentation aspect: declaring this version incompatible.

Issue #1825, opened 2022-08-30
Redis 5.0 Private Method Reconnect


PatrickTulskie:
Working through some of the issues with redis-rb 5.0... when running the tests with github, the CI runner would hang as soon as it started running specs. I did some debugging and found it was erroring out here:

```
/home/runner/work/resque/resque/lib/resque/server.rb:121: warning: instance variable @partial not initialized
/home/runner/work/resque/resque/lib/resque/server.rb:121: warning: instance variable @partial not initialized
/home/runner/work/resque/resque/lib/resque/server.rb:121: warning: instance variable @partial not initialized
/home/runner/work/resque/resque/lib/resque/data_store.rb:82:in `reconnect': private method `connect' called for #<Redis::Client redis://localhost:9736/0> (NoMethodError)
Did you mean?  connected?
	from /home/runner/work/resque/resque/test/child_killing_test.rb:27:in `block in start_worker'
	from /home/runner/work/resque/resque/test/child_killing_test.rb:25:in `fork'
	from /home/runner/work/resque/resque/test/child_killing_test.rb:25:in `start_worker'
	from /home/runner/work/resque/resque/test/child_killing_test.rb:81:in `block (2 levels) in <top (required)>'
```
The method reconnect in resque data_store calls connect again in the child after forking without closing the connection in the parent. For some reason that method is now private so we gotta come up with a work around.

DennisWormald:
Could this explain why this no longer works? I have the following code in my resque.rake task:

task "resque:pool:setup" do
  ActiveRecord::Base.connection.disconnect!
  Resque::Pool.after_prefork do |job|
    ActiveRecord::Base.establish_connection
    Resque.redis._client.reconnect
  end
end

casperisfine:
Yeah, that method was private API and is no longer usable in 5.x. 

I looked at it, and I'm not 100% sure what Resque was doing here. I think it does reconnect all the clients after fork, but just discard old connections without closing them to avoid impacting the parent.

Which is a bit weird to me because Ruby GC will automatically close connections when it collect them, but maybe `auto_close = false` was set somewhere. I'll have to dig into that.

Issue #1820, opened 2022-08-24
Redis 5.0.0 Returning Integer message


yilmaz-burak:
Hey,
I am getting
```
Redis#sadd will always return an Integer in Redis 5.0.0. Use Redis#sadd? instead.(called from: /var/lib/gems/2.7.0/gems/redis-namespace-1.9.0/lib/redis/namespace.rb:479:in `call_with_namespace')
```
when running worker and enqueuing. However, application works as expected.

Is that just me or a general thing? If so, is that important and do you plan to do something about it?

Love the product btw. Thanks!

burak-upstash:
Okay, at first i thought it was about Redis 6 vs Redis 5 kind of thing, but it seems that ruby-redis guys are about to release their v5...

yilmaz-burak:
Okay, at first i thought it was about Redis 6 vs Redis 5 kind of thing, but it seems that ruby-redis guys are about to release their v5...

PatrickTulskie:
Kind of a weird warning they are returning here. Resque isn't actually doing anything with the response, it just wants to add/remove from some sets so it will continue to function as expected.

In the interim if you want to silence the warnings, you can do this:
`Redis.sadd_returns_boolean = true`

I'll work through the code a bit and see if we need to apply a similar change, or if we need to change to the boolean method. It just feels un-ruby-like to add `?` to methods that are in fact making changes.

PatrickTulskie:
@yilmaz-burak what version of the redis gem are you using? 4.8.0?

yilmaz-burak:
Yeah, started getting these warnings when bumping to 4.8.0 from 4.7.1

ozzyaaron:
@PatrickTulskie AFAICT from what we see it only happens at startup. Unlike the [previous pipelining deprecation](https://github.com/resque/resque/pull/1817) which happened every call, this only seems to happen once upon Resque initialization.

PatrickTulskie:
Awesome thanks! Super helpful

PR #1828, opened 2022-08-31
redis-rb 5.0 compatiblity

casperisfine:
Fix: https://github.com/resque/resque/pull/1828
Fix: https://github.com/resque/resque/issues/1825
Fix: https://github.com/resque/resque/issues/1820
Close: https://github.com/resque/resque/pull/1823

There was two distinct issues.
    
The first one was that `Resque::Worker` instances were passed as argument to `sadd / srem` relying on the implicit call to `to_s`. This has been removed in redis-rb 5.0, you now must explictly cast to a String or another type understood by the redis client.
    
The other issue was with the reconnection after fork. When forking you must make sure to close the Redis connection, otherwise if you write in it, the responses of the parent and child will be mixed in.
    
Resque was using a private API to do this reconnection. I updated it to use the main public API.
    
There was also a bug in redis-rb 5.0.1 preventing from reconnecting after fork, I fixed it in 5.0.2 (https://github.com/redis/redis-rb/pull/1145).

NB: this is based on https://github.com/resque/resque/pull/1827 because the same lines are being edited so it avoid conflicts.

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -79,7 +79,9 @@ module Resque
     # Force a reconnect to Redis without closing the connection in the parent
     # process after a fork.
     def reconnect
-      @redis._client.connect
+      @redis.disconnect!
+      @redis.ping
+      nil
     end
 
     # Returns an array of all known Resque keys in Redis. Redis' KEYS operation
@@ -238,7 +240,7 @@ module Resque
 
       def register_worker(worker)
         @redis.pipelined do |piped|
-          piped.sadd(:workers, [worker])
+          piped.sadd(:workers, [worker.id])
           worker_started(worker, redis: piped)
         end
       end
@@ -249,7 +251,7 @@ module Resque
 
       def unregister_worker(worker, &block)
         @redis.pipelined do |piped|
-          piped.srem(:workers, [worker])
+          piped.srem(:workers, [worker.id])
           piped.del(redis_key_for_worker(worker))
           piped.del(redis_key_for_worker_start_time(worker))
           piped.hdel(HEARTBEAT_KEY, worker.to_s)
--- a/lib/resque/web_runner.rb
+++ b/lib/resque/web_runner.rb
@@ -286,7 +286,7 @@ module Resque
             begin
               handler = Rack::Handler.get(server)
               break
-            rescue LoadError, NameError => e
+            rescue LoadError, NameError
               next
             end
           end
--- a/test/child_killing_test.rb
+++ b/test/child_killing_test.rb
@@ -7,7 +7,7 @@ describe "Resque::Worker" do
     @queue = :long_running_job
 
     def self.perform( sleep_time, rescue_time=nil )
-      Resque.redis.reconnect # get its own connection
+      Resque.redis.disconnect! # get its own connection
       Resque.redis.rpush( 'sigterm-test:start', Process.pid )
       sleep sleep_time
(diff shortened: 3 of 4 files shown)


lib/resque/data_store.rb
```diff
@@ -76,10 +76,10 @@ def identifier
       @redis.inspect
     end
 
-    # Force a reconnect to Redis without closing the connection in the parent
-    # process after a fork.
-    def reconnect
-      @redis._client.connect
+    # Close the connection. Must be done after fork to avoid sharing the connection
+    # with the parent.
+    def disconnect!
```
PatrickTulskie, line 81: Is there any way to retain the reconnect functionality? Changing the method name on a public method might necessitate more of a version bump than I was hoping for right now. Additionally, it's not clear if after calling `disconnect!` if we're going to reconnect at a later time.

PatrickTulskie (changes requested):

PatrickTulskie:
I merged your other PR into master so if you rebase and update this PR, it will clean up the diff.

casperisfine:
Rebased.

lib/resque/data_store.rb
casperisfine: The Redis client always reconnected automatically.

If you want to force a reconnection, you can issue a command, e.g. `PING`.

casperisfine: I added `reconnect` back as an alias, it's functionally equivalent unless somehow you 100% rely on the connection being create right there and not on use.

lib/resque/worker.rb
```diff
@@ -347,22 +347,10 @@ def reserve
       raise e
     end
 
-    # Reconnect to Redis to avoid sharing a connection with the parent,
+    # Disconnect from Redis to avoid sharing a connection with the parent,
     # retry up to 3 times with increasing delay before giving up.
-    def reconnect
```
PatrickTulskie, line 352: Same here with this one since it's still technically a public method. If we can make an alias or an alternative method that would be ideal.

Also in this case, does it make sense to attempt the `PING` after disconnecting in order to retain the reconnect/retry logic? This is the only place resque seems to do this and I'm wondering if removing it might result in reduced fault tolerance. If it only disconnects and then we rely on the redis client to handle reconnecting, you may find connection errors at interesting new spots in your backtrace that you weren't expecting.

PatrickTulskie (changes requested):
Sorry to be a pain about changes on this PR... just trying to make sure that we're not impacting people who may have hooked into Resque and are expecting certain behaviors. I thought about it a lot over the course of the day and I think this is nearly there though.

lib/resque/worker.rb
casperisfine: > retain the reconnect/retry logic?

That reconnect/retry logic is totally redundant since Redis has it baked in, but ok. I honestly don't care, I don't use Resque.

casperisfine:
Done. I restored the behavior of `DataStore#reconnect` by using `ping`.

PatrickTulskie (approved):

lib/resque/worker.rb
PatrickTulskie: Yeah I hear ya... it was more out of concern for changing the behavior on public APIs since I know of other people using this that have various hooks and plugins that rely on this stuff. I appreciate you tweaking your PR though and I'm going to get a build going for early next week after Labor Day. Thanks!!

Issue #1787, opened 2022-01-19
Conflicting hooks between Resque and ActiveJob (e.g. before_enqueue)


hector:
I believe Resque hooks are clashing with ActiveJob ones.

Let 's see an example with the `before_enqueue` hook.
In a Resque job you can *define* the `before_enqueue` method as a hook to be executed before enqueueing; if false is returned the enqueuing does not happen ([docs](https://github.com/resque/resque/blob/master/docs/HOOKS.md#job-hooks)):
```ruby
class MyJob < ApplicationJob
  def before_enqueue
    # do something
  end
end
```

In ActiveJob, you can *use* the `before_enqueue` method to add a hook to be run before enqueueing ([api docs](https://api.rubyonrails.org/v7.0.1/classes/ActiveJob/Callbacks/ClassMethods.html#method-i-before_enqueue)):
```ruby
class VideoProcessJob < ActiveJob::Base
  queue_as :default

  before_enqueue do |job|
    $statsd.increment "enqueue-video-job.try"
  end

  def perform(video_id)
    Video.find(video_id).process
  end
end
``` 

Because of this clash:
1. Resque tries to run any `before_enqueue` method if it exists
2. It does always exist because it is defined by ActiveJob
3. Resque runs before_enqueue which makes no sense at all

The conflicting hooks are: after_enqueue, after_perform, around_enqueue, around_perform, before_enqueue, before_perform

hector:
I post the temporal workaround I am using:

```ruby
# Workaround for this problem: https://github.com/resque/resque/issues/1787
# Modify conflicting hooks (Resque vs ActiveJob) so that Resque hooks end with and underscore and
# so we can still use them as `before_enqueue_whatever`
module Resque
  module Plugin
    def before_hooks(job)
      get_hook_names(job, 'before_perform_')
    end

    def around_hooks(job)
      get_hook_names(job, 'around_perform_')
    end

    def after_hooks(job)
      get_hook_names(job, 'after_perform_')
    end

    def after_enqueue_hooks(job)
      get_hook_names(job, 'after_enqueue_')
    end

    def before_enqueue_hooks(job)
      get_hook_names(job, 'before_enqueue_')
    end
  end
end
```

iloveitaly:
PRs to fix this are welcome!

nvolker:
@iloveitaly, @hector I've put forward #1833 for discussion.

PR #1833, opened 2022-09-19
Drop support for unnamed job hooks

nvolker:
Solves: https://github.com/resque/resque/issues/1787
Related Documentation: https://github.com/resque/resque/blob/master/docs/HOOKS.md#job-hooks

Putting this pull request forward in the hopes to drum up some conversation around this issue. The linked issue encapsulates the problem this pull request aims to solve.

The solution on offer is opinionated and would likely result in a major release - this is not ideal. So, alternative suggestions are welcome!


PatrickTulskie:
I'm not against this, but I think we probably need a good way to add in some deprecation warnings before this gets merged in.

PR #1834, opened 2022-09-19
Escape `Resque.redis_id` for stats page

### LINKED ISSUE #1785: Stats page does not properly display redis server
`Resque.redis_id` used to return `Resque.redis.client.id`  which looks like `redis://127.0.0.1:6379/0`, but now it returns `Resque.redis.identifier` which looks like `<Redis::Namespace v1.7.0 with client v4.1.4 for redis://127.0.0.1:6379/0/resque>`

This causes the Redis server URL to not render properly in the `stats` page:

image

Is there any workaround for this?

I am using version 2.2.0.

Thanks!

mishina2228:
Fix a part of #1785
Refs #1832

This PR escapes `Resque.redis_id` to properly display information of redis server.

After this change:
- resque tab
  stats2
- redis tab
  redis_tab
- keys tab
  keys_tab

It includes extra stuff other than connection information, but I think that more information is better, so I just escaped the value.

### DIFF
--- a/lib/resque/server/views/layout.erb
+++ b/lib/resque/server/views/layout.erb
@@ -38,7 +38,7 @@
 
 <div id="footer">
   <p>Powered by <a href="http://github.com/resque/resque">Resque</a> v<%=Resque::VERSION%></p>
-  <p>Connected to Redis namespace <%= Resque.redis.namespace %> on <%=Resque.redis_id%></p>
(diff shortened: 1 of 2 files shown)


mishina2228:
I noticed that there were other tabs a few days ago, so I made some additional corrections.

PR #1838, opened 2022-09-24
Show correctly the values of hash and none type on stats tab

mishina2228:
The keys and values of Redis can be checked in Resque Stats tab.
[resque-scheduler](https://github.com/resque/resque-scheduler) registers hash type value on Redis for persistent schedules data.
However, when I try to display the value, it raises an error.

- `persistent_schedules` is a hash. Its size is not shown.
before_stats1
- Error after clicking the key
before_stats2

This PR will solve this problem.
- `persistent_schedules`' size is shown.
after_stats1
- After clicking, we can see the value.
after_stats2

In addition, this PR corrected the case where the type is none.
`#redis_get_size` used to return an empty array(`[]`), but it should return `0` here.

### DIFF
--- a/lib/resque/server.rb
+++ b/lib/resque/server.rb
@@ -67,7 +67,9 @@ module Resque
       def redis_get_size(key)
         case Resque.redis.type(key)
         when 'none'
-          []
+          0
+        when 'hash'
+          Resque.redis.hlen(key)
         when 'list'
           Resque.redis.llen(key)
         when 'set'
@@ -83,6 +85,8 @@ module Resque
         case Resque.redis.type(key)
         when 'none'
           []
+        when 'hash'
+          Resque.redis.hgetall(key).to_a[start..(start + 20)]
         when 'list'
           Resque.redis.lrange(key, start, start + 20)
         when 'set'


lib/resque/server.rb
```diff
@@ -83,6 +85,8 @@ def redis_get_value_as_array(key, start=0)
         case Resque.redis.type(key)
         when 'none'
           []
+        when 'hash'
+          Resque.redis.hgetall(key).to_a[start..(start + 20)]
```
mishina2228, line 89: `HGETALL` has a noticeable performance problem when the size of the hash is large.
We do not expect to use such large hashes in Resque, so I use `HGETALL` here.

PatrickTulskie:
@mishina2228 hey just going through things here getting caught up. I know it's been a while, but do you think you can rebase your branch and add a test for your change?

PatrickTulskie:
> I'm trying to add tests but it could be tough. sweat_smile To test the methods in the `helpers` block, it appears that the code in the block needs to be cut out into a module. https://stackoverflow.com/questions/2878818/how-can-i-test-helpers-blocks-in-sinatra-using-rspec

Ah okay maybe not worth it in that case then. I think we should be good as-is.

PatrickTulskie (approved):

mishina2228:
Thanks! :bow:

PR #1846, opened 2022-12-06
Replace `File.exists?` with `File.exist?`

kyoshidajp:
`File.exists?` will be removed in Ruby 3.2.

https://www.ruby-lang.org/en/news/2022/11/11/ruby-3-2-0-preview3-released/


PatrickTulskie (approved):

PatrickTulskie:
Nice LGTM. Not sure when the next release will be, but this will be in it.

PR #1850, opened 2022-12-29
Update CI matrix

mishina2228:
- Add Ruby 3.2 to the CI matrix
- Enclose 4.0 in quotes
  - Without quotes, `4.0` is interpreted as `4`.
    - Before: 
image
    - After: 
image


PatrickTulskie:
Ah thanks was meaning to do that.

PR #1879, opened 2023-08-18
Fix MiniTest undefined errors

PatrickTulskie:
Not sure why this changed, but apparently MiniTest is actually Minitest. Only recently noticed this start crashing in CI.


corincerami:
Yeah, the newest version of Minitest finally removed the very old MiniTest compatibility namespace.

PR #1880, opened 2023-08-18
Fix failing reconnect specs

PatrickTulskie:
Test that we call Redis#disconnect! at the right time instead of testing the `object_id` of the socket after forking.

Changes in redis-client 0.14.0 make it so that the `object_id` persists even after reconnecting.

Update:
Actually just test that redis connections are available after forking. Reasoning in the comments below.

### DIFF
--- a/test/worker_test.rb
+++ b/test/worker_test.rb
@@ -1266,11 +1266,10 @@ describe "Resque::Worker" do
   end
 
   it "no disconnect from redis when not forking" do
-    original_connection = redis_socket(Resque.redis)
     without_forking do
+      assert_equal Resque.redis.ping, "PONG"
       @worker.work(0)
     end
-    assert_equal original_connection, redis_socket(Resque.redis)
   end
 
   it "logs errors with the correct logging level" do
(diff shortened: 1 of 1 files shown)


PatrickTulskie:
This is the changelog entry that tipped me off: https://github.com/redis-rb/redis-client/commit/db646d1a25aea263eab3f146b76c2b533bbc77f7#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR4

@byroot if you have another suggestion on how to fix this test I'm open to changing this.

byroot:
> if you have another suggestion on how to fix this test I'm open to changing this.

Well, thing is you don't need to explicitly disconnect with semi-recent `redis` / `redis-client`. 

`resque` requires `redis-namespace` which requires `redis >= 4.x`, so  I think this code is simply no longer required.

You can simply test that redis still work after a fork.

PatrickTulskie:
@byroot that's a good point. I just updated it so that it simply checks Redis with a `ping` after forking (and when not forking) in order to satisfy the original spirit of the tests. Lemme know what you think.

Since redis 4.x automatically reconnects after forking, next steps are to deprecate the resque reconnect methods for people calling it explicitly, and remove the reconnect logic from forking.

I want to do this as gently as possible since there's a lot of old code out there using resque, some of it using undocumented APIs from blog entries and incantations passed down from generations of developers. I'd rather people encounter deprecation warnings and NOOPs than having their background job processing break or their apps not boot.

byroot:
> Lemme know what you think.

LGTM.

> Since redis 4.x automatically reconnects after forking

Would be good to double check that is true of 4.0.0, I don't know for sure when the automatic reconnect was introduced exactly.

> next steps are to deprecate the resque reconnect methods

I don't think that's necessary. I can't really think of why, but if people wish to reconnect for some reason, why not let them do so?

PatrickTulskie:
> Would be good to double check that is true of 4.0.0, I don't know for sure when the automatic reconnect was introduced exactly.

Oh okay. Good call. I'll figure out which version and then up the minimum version requirement.

> > next steps are to deprecate the resque reconnect methods
> 
> I don't think that's necessary. I can't really think of why, but if people wish to reconnect for some reason, why not let them do so?

True. I could just leave them in and then not call them when forking. I guess if you need to reconnect to redis within a worker for some reason, having access to it there in the context of a worker is handy.

Thanks!

PR #1886, opened 2023-11-17
add always hooks

jarednielson:
Adds 'always' hooks which run in an ensure block in perform.

We found this feature useful for certain clean up tasks that we always want to run.
Often we would follow the pattern

```rb
def self.on_failure_zzz_cleanup(exception, *args)
  cleanup!
end

def self.after_perform_zzz_cleanup(*args)
  cleanup!
end
```

With an always hook this can be simplified to

```rb
def self.always_cleanup(*_args)
  cleanup!
end
```

There are test cases around what error messages are propagated when:
Just the code in the always hook raises an error
The code in perform and an always hook raise errors
The code in perform, on_failure hooks and always hooks raise errors

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,9 @@
+## 2.7.0
+
+### Added
+
+* `always` hooks that run regardless of the success or failure of the job(#1886)
+
 ## 2.6.0
 
 ### Fixed
--- a/docs/HOOKS.md
+++ b/docs/HOOKS.md
@@ -127,6 +127,9 @@ The available hooks are:
 * `on_failure`: Called with the exception and job args if any exception occurs
   while performing the job (or hooks), this includes Resque::DirtyExit.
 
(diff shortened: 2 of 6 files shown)


PatrickTulskie:
Hey so we took another run at this and due to the merge conflicts we couldn't bring it in exactly. Instead we picked it into another PR, leaving commit attribution in place, and then addressed a few issues and updated docs for this. I'm going to close this out as superseded by #1966 thanks!!!

jarednielson:
@PatrickTulskie Wow! Thanks for doing the work to get it picked over to a clean branch and merged in. Cheers

PatrickTulskie:
> @PatrickTulskie Wow! Thanks for doing the work to get it picked over to a clean branch and merged in. Cheers

No problem! Thanks for the addition. Sorry it took so long to get it in.

PR #1887, opened 2023-11-21
Remove obsolete enqueue_front option

alxckn:
b45a6852fe7139286fc8f2733c6b4753e5a7e463 had introduced a `enqueue_front` option which enabled LIFO job execution instead of FIFO.

But a refactoring in 4bb44413a045fca78d25e22ad535b3f5ea97fab8 ignored that change.

This is a proposal to remove this unused property to avoid confusion about it not being functional.

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -207,19 +207,6 @@ module Resque
     end
   end
 
-  # By default, jobs are pushed to the back of the queue and popped from
-  # the front, resulting in "first in, first out" (FIFO) execution order.
-  # Set to true to push jobs to the front of the queue instead, resulting
-  # in "last in, first out" (LIFO) execution order.
-  attr_writer :enqueue_front
-  def enqueue_front
-    if defined? @enqueue_front
-      @enqueue_front
-    else
-      @enqueue_front = false
-    end
-  end
-
   # The `before_first_fork` hook will be run in the **parent** process
   # only once, before forking to run the first job. Be careful- any
   # changes you make will be permanent for the lifespan of the


PatrickTulskie:
@alxckn do you think maybe it makes sense to just fix the functionality instead of deleting the property?

alxckn:
@PatrickTulskie I think this functionality was active in the master branch only for a couple of months 7 years ago.

So there does not seem to be any use for it at the moment. At least not in this way where the whole application is configured to use LIFO over FIFO.
I know some comments mentioned this type of functionality could be useful at a queue level or even at a "per enqueue" level (I could be interested in contributing on this aspect). 

Which is why I believe that the feature as it is is better removed to avoid confusion when reading the code docs.

PatrickTulskie:
If it is getting removed, since the method has been around for a while, it's probably best to deprecate it properly and then remove it in the next major release.

That said, since the method is there and it's possible people out there are making calls to it and are expecting it to work, it might be better to just fix the bug. It does seem like a useful feature.

alxckn:
You're probably right about being more careful to deprecate the method in steps.

Your comment makes me think of another argument against fixing this issue. If I had activated this option without noticing it does nothing, I wouldn't want it to become functional again all of a sudden and have my prioritisation strategy change and have strange behaviors in production. 

WDYT?

PatrickTulskie:
@alxckn sorry I missed your response before... I think if we just fix it and put it in the release notes along with semver, it shouldn't be too surprising if this works again. Still seems like there might be a need for it.

PR #1891, opened 2023-12-05
Update README.markdown

danlo:
Added note about dependency requirement of rack-session 2.0.0 using Rack 3, which sinatra does not support.  Suggesting to user that they downgrade to rack-session 1.0.2

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -133,6 +133,15 @@ require 'your/app' # Include this line if you want your workers to have access t
 
 #### Rails
 
+##### Rails 7.1
+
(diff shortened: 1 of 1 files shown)


pboling (approved):
LGTM

mishina2228:
@danlo [Sinatra 4.0](https://github.com/sinatra/sinatra/blob/main/CHANGELOG.md#400--2024-01-19) supports Rack 3.0, so we can use rack-session 2.0 with Rails 7.1.

pboling:
@mishina2228 Excellent!  PRs welcome, otherwise, I'll try to get around to re-updating the README.

PR #1903, opened 2024-07-30
docs: cleanup

dijonkitchen:
Doesn't seem like this is relevant anymore.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -135,13 +135,6 @@ require 'your/app' # Include this line if you want your workers to have access t
 
 ##### Rails 7.1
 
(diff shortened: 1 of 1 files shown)


PatrickTulskie:
@dijonkitchen did this get resolved? I haven't been keeping up with the compatibility issue there.

pboling:
@dijonkitchen @PatrickTulskie Yes, the fix was integrated here: https://github.com/sinatra/sinatra/pull/1857
So with the release of Sinatra 4.0 this is no longer needed.

README.markdown
```diff
@@ -135,13 +135,6 @@ require 'your/app' # Include this line if you want your workers to have access t
 
 ##### Rails 7.1
```
zzak, line 137: @dijonkitchen I think this "Rails 7.1" subheading is no longer needed too if you want to send a PR. :bow:

pboling: Indeed, that makes sense.

PR #1907, opened 2024-10-16
Ensure ruby 3.3 compatibility

xjunior:
MonoLogger adds ruby 3.3 compatibility on 1.1.2

### DIFF
--- a/resque.gemspec
+++ b/resque.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_dependency "redis-namespace", "~> 1.6"
   s.add_dependency "sinatra", ">= 0.9.2"
   s.add_dependency "multi_json", "~> 1.0"
-  s.add_dependency "mono_logger", "~> 1.1.2"
+  s.add_dependency "mono_logger", "~> 1"
 
   s.description = <<description
     Resque is a Redis-backed Ruby library for creating background jobs,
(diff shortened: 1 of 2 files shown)


resque.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_dependency "redis-namespace", "~> 1.6"
   s.add_dependency "sinatra", ">= 0.9.2"
   s.add_dependency "multi_json", "~> 1.0"
-  s.add_dependency "mono_logger", "~> 1.0"
+  s.add_dependency "mono_logger", "~> 1.1.2"
```
santiagorodriguez96, line 28: The current version of `resque` supports installing `mono_logger` `v1.1.2`, which adds Ruby 3.3 support. 

However, by constraining the gem to a patch level, we would be limiting users to a specific minor version of `mono_logger` v1, increasing friction when newer minor versions of `mono_logger` are released. It would also mean that users might have to update `mono_logger` unnecessarily if they want to update `resque` if they are not using Ruby 3.3.

In my opinion, providing the flexibility of installing whatever minor version of `mono_logger` `v1` that the users want – which means that users of the gem that are running it with Ruby 3.3 should upgrade `mono_logger` to `v1.1.2` – would be a better approach here.

PatrickTulskie (approved):

PR #1912, opened 2024-12-28
Rack 3.x support

and9000:
Hello, 

The goal of this PR is to make `resque` support `rack ~> 3` (so `sinatra ~> 4`) or at least start the conversation 😄 

`resque` ATM depends on `rack ~> 2` as its dependency `thin` depends on it (there's a [PR](https://github.com/macournoyer/thin/pull/399) on the topic but it has been open since 1 year). On this PR, I removed `thin` if favor of `puma` to remove that blocker. 

As `rack >= 3` moved handlers to `rackup` I've added it as a dev dependency but I had to drop support for `ruby 2.3` as `rackup` requires `ruby >=2.4`.

`Gemfile` has been modified so different ruby versions use different `rack` versions - there are old versions of some dependency which does specify correctly which version of `rack` it requires.

Code is backward compatible with `rack < 3` so when updating `resque` there's no need to update also `rack/sinatra/etc.`, I only had to drop `ruby 2.3` support (EOL: 2019-03-31).

What do you think? Do you already have a plan to add support to `rack >= 3`?

Thanks,
Andrea

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,6 @@ jobs:
           - 4.8
           - latest
         ruby-version:
-          - 2.3
           - 2.4
           - 2.5
           - 2.6
@@ -34,11 +33,6 @@ jobs:
           - 3.2
           - head
         exclude:
-          - { ruby-version: 2.3, redis-version: 4.5 }
-          - { ruby-version: 2.3, redis-version: 4.6 }
-          - { ruby-version: 2.3, redis-version: 4.7 }
-          - { ruby-version: 2.3, redis-version: 4.8 }
-          - { ruby-version: 2.3, redis-version: latest }
           - { ruby-version: 2.4, redis-version: latest }
     env:
       REDIS_VERSION: "${{ matrix.redis-version }}"
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,16 @@ end
 gem "json"
 gem "minitest", "~> 5.11"
 gem "mocha", "~> 2.0", require: false
+
+ruby_version = Gem::Version.new(RUBY_VERSION)
+if ruby_version >= Gem::Version.new("2.4") && ruby_version < Gem::Version.new("2.6")
+  gem "rack", "~> 1"
+elsif ruby_version >= Gem::Version.new("2.6") && ruby_version < Gem::Version.new("2.7")
+  gem "rack", "~> 2"
+else
+  gem "rack"
+end
+
 gem "rack-test", "~> 2.0"
 gem "rake"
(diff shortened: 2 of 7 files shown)


PatrickTulskie:
Thanks for putting this together. I actually was doing a similar PR for resque-scheduler just now and found some similar issues. Supporting unsupported versions of Ruby is definitely a problem at this point so I'm thinking that it's time for a major version bump and a clean break.

Anyhow, I'm going to prep a final minor version bump to cover the latest updates and then I'll start putting together pre release updates for resque, resque-scheduler that drops support for unsupported versions of Ruby. At that point I'll merge this in and some other things that have been lingering.

and9000:
Great, thanks! If you have a plan and want to share/split the work I could help both on resque and resque-scheduler, as a matter of fact I've an open [PR](https://github.com/resque/resque-scheduler/pull/715) on resque-scheduler which I could bring back to life once everything is green

PatrickTulskie:
@and9000 Appreciate it. I'm working up a little bit of a list now and I'll probably put together some issues. Most of the things I think need doing already have some preliminary PRs which is good. Gonna try to start setting up the repos for the next major versions this week and organize my thoughts a bit.

and9000:
Hi @PatrickTulskie , support for rack 3 has been merged in thin https://github.com/macournoyer/thin/pull/399 , what do you think if I resume this PR - removing the switch from Thin to Puma as it is not necessary anymore - and make `resque` support rack 3? 😄

PatrickTulskie:
@and9000 hey sorry for the delay here... I was thinking it's about time to prepare for a Resque 3.x release and I'd like to use this PR as the base for that.

With that in mind, would you do 2 things for me?

1. Create a separate commit for just the removal of ruby versions from CI
2. Squash down the rest of the commits into 1 about supporting Rack 3.

After that, I think we can merge this in and I'll do a version 3 release.

and9000:
Hi @PatrickTulskie , sure! In the next couple of weeks I should have some time, I'll open a PR to remove old ruby version the rebase and squash this one.

and9000:
Hi @PatrickTulskie , I've updated the PR following your instructions.

The only 2 things I had to do more are:
* bump redis gem `gem 'redis', '~> 5.0', '< 5.4.1'` as `5.4.1` seems to have an [issue](https://github.com/redis/redis-rb/issues/1321) - or better a dependency seems not loading it properly. I'll follow the issue and try to understand what's causing it
* @zzak I dropped CI code to test activejob adapter for rails 7.1 as `ActiveJob::QueueAdapters::AbstractAdapter` has been introduced in rails 7.2

PR #1914, opened 2025-04-10
feat: add `link` and `script` tag helpers

### LINKED ISSUE #1897: Overview UI Broken Due to Nonce Whitelist
Hi all, I hope this finds you well!

The security requirements for our app dictate that we must whitelist all JavaScript via the [nonce attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/nonce).

This breaks the "Live Poll" feature of Resque's UI (resque/overview), as it doesn't update or have CSS. I've looked into the documentation and issues here but have not found a solution.

Is there a way to add the nonce or edit the view? Or how can we get around this?

Thank you

SimenB:
These can be overridden by consumers to add e.g. `nonce`.

In our `config/resque.rb` I've added the following

```ruby
module Resque
  module ServerHelper
    def script_tag(src)
      base = "<script src=\"#{url_path(src)}\" type=\"text/javascript\"></script>"

      request = ActionDispatch::Request.new(env)

      nonce = request.content_security_policy_nonce

      return base unless nonce

      base.gsub(/><\/script>$/, " nonce=\"#{nonce}\"></script>")
    end

    def link_tag(src)
      base = "<link href=\"#{url_path(src)}\" media=\"screen\" rel=\"stylesheet\" type=\"text/css\">"

      request = ActionDispatch::Request.new(env)

      nonce = request.content_security_policy_nonce

      return base unless nonce

      base.gsub(/">$/, "\" nonce=\"#{nonce}\">")
    end
  end
end
```

And with that, the CSP errors are gone 🥳 

image

(Note that I am by no means a ruby/rails/sinatra/rack expert, but this _works_. Happy to change approach if there are better ways to go about this. I'd like to avoid having to maintain a fork, tho)

Fixes #1897

### DIFF
--- a/lib/resque/server_helper.rb
+++ b/lib/resque/server_helper.rb
@@ -19,7 +19,7 @@ module Resque
     alias_method :u, :url_path
 
     def script_tag(src)
-      "<script src=\"#{url_path(src)}\" type=\"text/javascript\"></script>"
+      %Q(<script src="#{url_path(src)}" type="text/javascript"></script>)
     end
 
     def link_tag(src)
--- a/lib/resque/server_helper.rb
+++ b/lib/resque/server_helper.rb
@@ -23,7 +23,7 @@ module Resque
     end
 
     def link_tag(src)
-      "<link href=\"#{url_path(src)}\" media=\"screen\" rel=\"stylesheet\" type=\"text/css\">"
+      %Q(<link href="#{url_path(src)}" media="screen" rel="stylesheet" type="text/css">)
     end
 
     def path_prefix
--- a/lib/resque/server_helper.rb
+++ b/lib/resque/server_helper.rb
@@ -18,6 +18,14 @@ module Resque
     end
     alias_method :u, :url_path
 
+    def script_tag(src)
+      "<script src=\"#{url_path(src)}\" type=\"text/javascript\"></script>"
+    end
+
+    def link_tag(src)
+      "<link href=\"#{url_path(src)}\" media=\"screen\" rel=\"stylesheet\" type=\"text/css\">"
+    end
+
     def path_prefix
       request.env['SCRIPT_NAME']
     end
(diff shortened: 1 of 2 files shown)


lib/resque/server_helper.rb
```diff
@@ -18,6 +18,14 @@ def url_path(*path_parts)
     end
     alias_method :u, :url_path
 
+    def script_tag(src)
+      "<script src=\"#{url_path(src)}\" type=\"text/javascript\"></script>"
```
PatrickTulskie:
```suggestion
      %Q(<script src="#{url_path(src)}" type="text/javascript"></script>)
```

lib/resque/server_helper.rb
```diff
@@ -18,6 +18,14 @@ def url_path(*path_parts)
...
+    end
+
+    def link_tag(src)
+      "<link href=\"#{url_path(src)}\" media=\"screen\" rel=\"stylesheet\" type=\"text/css\">"
```
PatrickTulskie:
```suggestion
      %Q(<link href="#{url_path(src)}" media="screen" rel="stylesheet" type="text/css">)
```

PatrickTulskie (changes requested):
I think this is okay. Just two formatting suggestions to make things a little easier to read.

If you can update your PR in the next day, I should be able to sneak this into 3.0.

lib/resque/server_helper.rb
SimenB: oh cool, didn't know about this syntax!

SimenB:
@PatrickTulskie I updated btw - dunno if an email fires when I accept suggestions 😅

SimenB:
@PatrickTulskie hey any chance this can land now that v3 is out the door? 🙂

PatrickTulskie (approved):

PR #1916, opened 2025-06-09
Add shutdown hooks

djmb:
Rails is getting a new feature called
[Active Job Continuations](https://github.com/rails/rails/pull/55127), which will allow jobs to be interrupted and resumed.

For this to be integrated with Resque, we need to know when a worker is shutting down.

There's a new `stopping?` method on the Rails queue adapter. Add a `shutdown` hook to Resque to allow us to implement that method.

### DIFF
--- a/lib/resque.rb
+++ b/lib/resque.rb
@@ -316,6 +316,20 @@ module Resque
     register_hook(:worker_exit, block)
   end
 
+  # The `shutdown` hook will be run in the **child** process
+  # when the worker has received a signal to stop processing
+  #
+  # Call with a block to register a hook.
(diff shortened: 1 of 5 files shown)


lib/resque.rb
```diff
@@ -316,6 +316,20 @@ def worker_exit=(block)
     register_hook(:worker_exit, block)
   end
 
+  # The `shutdown` hook will be run in the **child** process
```
bquorning, line 319: docs/HOOKS.md says this hook will run in the **parent** process, which I think is correct.

bquorning: Fixing this, and adding test coverage in #1932.

PR #1918, opened 2025-09-11
Fix re-queued job timezone data

luizbettoni:
#### Origin

At our timezone (`-03`), immediately after retry a job on _reque web_, the `retried_at` information displayed is wrong:

> image

> image

#### Fix

Add timezone information to the field.
Seems that `retried_at` deserves same respect as `failed_at`.


PatrickTulskie (approved):
LGTM. Once I get the CI stuff sorted out I'll tee this up for merging in too.

PR #1919, opened 2025-09-24
Fix a double redis.get to pause-all-workers in each work loop

mdkent:
While debugging some performance issues with a huge volume of workers I noticed that 
```
redis.get('pause-all-workers')
```
is being run twice for each interval. If you set the INTERVAL to 100ms and run a thousand workers, this adds up!

This change caches the paused state and uses it to set the procline.

Here's the impact for us:

Screenshot 2025-09-24 at 11 15 34 PM

### DIFF
--- a/lib/resque/worker.rb
+++ b/lib/resque/worker.rb
@@ -143,6 +143,7 @@ module Resque
     def initialize(*queues)
       @shutdown = nil
       @paused = nil
+      @cached_pause_value = nil
       @before_first_fork_hook_ran = false
 
       @heartbeat_thread = nil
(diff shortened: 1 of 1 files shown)


nevans (approved):
LGTM!

mdkent:
> @mdkent hey I'm not sure if I'm missing something here, but doesn't this change just disable the `pause-all-workers` functionality? The value is cached the first time `paused?` is checked, and then it just never checks it again during the work loop right?

Oh! `@paused` will be false in the `work_one_job` loop, so it's constantly running the `||` condition to `redis.get('pause-all-workers').to_s.strip.downcase == 'true'`

nevans:
> The value is cached the first time `paused?` is checked, and then it just never checks it again during the work loop right?

Maybe I'm confused or missing something... but the cache var is assigned with `=` not `||=`, so it shouldn't change the behavior of `#pause?` at all and should be updated every time `#paused?` is called explicitly.  And `#paused?` is still called explicitly where it matters, at the beginning of `#work_one_job`.  This cache should only affect the procline, right?

Maybe the variable should be renamed to reflect that this is just a cached result?  Maybe `@cached_pause_state`?

(A few months ago, I fixed a similar bug in a different daemon: the procline was being updated by running slow code _which required thread synchronization_.  So the procline code was causing huge slowdowns because the "actual" code had to wait for locks so often.)

PR #1920, opened 2025-09-30
Support a dynamic sleep INTERVAL

mdkent:
We run a large resque (resque-pool) installation with ~1100 workers per site. These use a low INTERVAL of 100ms to pickup new work. Something we've noted is that the workers with empty queues pummel the redis server endlessly trying to find the next batch of work. This seems a bit wasteful on the CPU side, and is impacting the quick pickup by workers that are actually busy.

This change adds a dynamic sleep interval that backs off to a maximum value. This gives us the quick pickup we need, while the more idle workers can chill a bit.

This cut a good 15k/sec of redis commands:
Screenshot 2025-09-26 at 2 42 44 PM

and 20Mbit/s of network traffic
Screenshot 2025-09-26 at 2 43 26 PM

See https://github.com/resque/resque-pool/pull/251 for the resque-pool side.

### DIFF
--- a/lib/resque/tasks.rb
+++ b/lib/resque/tasks.rb
@@ -17,7 +17,11 @@ namespace :resque do
 
     worker.prepare
     worker.log "Starting worker #{worker}"
-    worker.work(ENV['INTERVAL'] || ENV['MAX_INTERVAL'] || 5, ENV['MIN_INTERVAL'] || 5, ENV['BACKOFF_INTERVAL'] || 0.1) # interval, will block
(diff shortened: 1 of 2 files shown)


mdkent:
Okay, I incorporated the feedback - all of it verbatim.

I've rested this branch on our staging setup and it looks good.

PatrickTulskie:
@nevans sounds good yeah definitely let me know. I was thinking about trying to get this and #1919 into 3.0, but might have to wait for 3.1 depending on timing of things.

PR #1925, opened 2025-12-10
Some nice fixes 

taf2:
Some nice fixes to make resque web slightly nicer and avoid redis del when we can instead use unlink.

### DIFF
--- a/lib/resque/data_store.rb
+++ b/lib/resque/data_store.rb
@@ -133,7 +133,7 @@ module Resque
       def remove_queue(queue)
         @redis.pipelined do |piped|
           piped.srem(:queues, [queue.to_s])
-          piped.del(redis_key_for_queue(queue))
+          piped.unlink(redis_key_for_queue(queue))
         end
       end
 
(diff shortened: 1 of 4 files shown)


taf2:
@nevans thanks and sorry i actually was trying to rebase this and pr into our repo - but if we can merge these upstream that'd be amazing and i'll push a new pr for each feature.

PatrickTulskie (changes requested):
@taf2 I was actually going to do something similar to your `unlink` commit in a separate branch I was working on when I stumbled upon your PR here so I gave it a quick review.

Re: Auto start of polling - I probably wouldn't add that to the gem since during triage where you have several engineers looking at things, it's possible you can have a thundering herd problem. Better to just open the page and let people turn on polling if they want to.

I think the other changes warrant their own PRs though. Can you carve them off into one for the dark mode and one for the unlinking? For the dark mode one, screenshots would be great too.

If you don't have time or aren't interested in contributing to resque anymore I totally understand just lemme know.

PatrickTulskie:
Closing this out. We actually picked in parts of this PR with #1962 and #1963 and addressed a few issues that were preventing them from being merged. Thanks!!

PR #1932, opened 2026-03-31
Test which hooks run in which process

bquorning:
Added to test/resque_hook_test.rb a test for each hook with an assertion about whether it is executed in the parent or the child process.

Fixed a code comment claiming that the the "shutdown" hook is run in the child process, when it is actually run in the parent process. The documentation in HOOKS.md was correct though.

Added documentation about the "before_pause" and "after_pause" hooks to HOOKS.md


PatrickTulskie (approved):
Thanks for sorting this out!

PR #1941, opened 2026-07-15
Regression tests for the failure-drop bug + 1.27.4.2

davisonja:
Follow-up to f4456be (the `report_failed_job` fix, verified live on ps-queue4 today): regression tests so this class of bug can't ship silently again, plus the version bump so the fixed code self-identifies (broken 1.27.4.1 is currently deployed; `ps | grep resque-1.27.4.2` will mark rollout progress host by host).

- **Tests**: a failing job must reach the failure backend; the per-host/per-queue `failed:` stat family (the fork's raison d'être) must move; the no-job fallback works. All exercise the real `Resque::Job` call shape that crashed.
- **Harness housekeeping** (first local run in years, evidently): `.ruby-version` 3.3.6, production-parity redis pin (5.x breaks 1.27's client construction), CI-only coveralls, `mocha/minitest`.
- Full suite: 218 tests — only pre-existing failures are `resque-web_test.rb` (sinatra/rack dep drift, untouched here).

Pairs with the pocketsmith lock bump `2dd1f3d` → this HEAD, which is the change that actually fixes queue0–3.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved failure tracking with accurate global, host-specific, queue-specific, and host-and-queue statistics.
  - Ensured failed jobs are recorded even when job details are unavailable.
  - Improved processed-job statistics by queue and host.

- **Chores**
  - Updated Redis compatibility requirements.
  - Updated the release version to 1.27.4.2.

- **Documentation**
  - Added an Unreleased section to the changelog.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,11 @@ source "https://rubygems.org"
 
 gemspec
 
+# production parity (pocketsmith Gemfile.lock): redis 5.x breaks resque 1.27's
+# client construction, and tests should exercise what production runs anyway.
+gem 'redis', '~> 4.2.0'
+gem 'redis-namespace', '~> 1.11.0'
+
 gem "activesupport", "~> 3.0"
 gem "i18n"
 
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+Nothing yet!
+
+## 1.27.4 (2017-4-15)
+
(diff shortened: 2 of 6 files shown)


davisonja:
Apologies — opened against upstream by mistake (fork-default base); this belongs on our fork.

